PR #35, opened 2024-04-01
Fix: Using Image tag instead of img for user avatar in tweets.

its-kunal:
This pull request fixes the issue - [Link](https://github.com/keploy/keploy/issues/1762)

- use of an Image tag will leverage the full potential of Next.js.

### DIFF
--- a/components/tweets.tsx
+++ b/components/tweets.tsx
@@ -11,10 +11,12 @@ const Tweets = ({ avatar, name, id, post, content }) => {
       >
         <div className="flex flex-row items-center justify-between">
           <div className="flex flex-row gap-2 items-center">
-            <img
+            <Image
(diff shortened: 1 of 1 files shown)


components/tweets.tsx
```diff
@@ -11,10 +11,12 @@ const Tweets = ({ avatar, name, id, post, content }) => {
       >
         <div className="flex flex-row items-center justify-between">
           <div className="flex flex-row gap-2 items-center">
-            <img
+            <Image
               src={avatar}
               alt="profile_image"
-              className="rounded-full h-12"
+              width={"100"}
```
Hermione2408, line 17: @its-kunal you have added width twice I don't think it is necessary

PR #36, opened 2024-04-01
Fix: use Image tag instead of img tag in open source vector png.

its-kunal:
This pull request fixes the issue - [Link](https://github.com/keploy/keploy/issues/1763)

The use of `Image` tag will leverage the use of Next.js framework.


nehagup:
@its-kunal please ensure that it builds for node 20, we're planning to migrate.

its-kunal:
@nehagup, yes the current changes support the build for node 20.

PR #34, opened 2024-03-30
Add animations to blog website

sahadat-sk:
Fixing: [#1664](https://github.com/keploy/keploy/issues/1664)

**Features added**

- Added loading screen animation
- Hover button animation on the 'technology' and 'community' buttons in the landing page
- Hover animation on blog preview

https://github.com/keploy/keploy/assets/98450585/759527c3-b293-4074-8942-6aa781887403

PR #37, opened 2024-04-01
Fix: use of Link tag instead of anchor tag.

its-kunal:
This pull request fixes the issue - [Link](https://github.com/keploy/keploy/issues/1764)
Also, I know there is still the use of img tags in the code. but I have covered them in the pull request - #34


Hermione2408 (commented):
locally ran build works fine, LGTM @nehagup

PR #56, opened 2024-10-08
fix: author profiles

Sonichigo:
tested locally with `npm run build`

image

In this PR, we have update `postAuthor` query back to `ppmaAuthorName` and disabled yoast plugin. Since PublishPressAuthors and Yoast are in compatible. We explore Mologui Plugin with PostAuthor, it worked best, but at many places certain values were hardcoded, leading to breaking changes in author data.


khareyash05:
GTG! @Sonichigo

khareyash05 (approved):

PR #89, opened 2025-02-12
fix: placed json-diff-viewer at top

SaketV8:
Placed `json-diff-viewer` just below the header banner to make it easier for readers to find.

### Fixes (issue)
https://github.com/keploy/keploy/issues/2524

### After | Before

keploy-blog-after-before


Hermione2408:
LGTM!

PR #99, opened 2025-02-22
feat: implement pagination and search functionality in MoreStories co…

MishraShardendu22:
Fixes issue : https://github.com/keploy/keploy/issues/2546

Implemented Pagination for Blogs Pages. Points Kept in mind 
1.) UI of the site matches the newly added component 
2.) The number of pages visible at a time are limited so the bottom of the pages aren't  flooded.

image


manasmanohar:
Thank you for your implementation!
Your approach works well for controlling what's displayed, but doesn't address the underlying issue. Currently, all posts are still fetched initially, and then just displayed incrementally. A better approach would be, to load data incrementally - fetching only a subset of posts from the API initially, then requesting more as needed upon user clicking load more.

MishraShardendu22:
> Thank you for your implementation! Your approach works well for controlling what's displayed, but doesn't address the underlying issue. Currently, all posts are still fetched initially, and then just displayed incrementally. A better approach would be, to load data incrementally - fetching only a subset of posts from the API initially, then requesting more as needed upon user clicking load more.

Ok Sir I'll address the issue and fix it by today

MishraShardendu22:
I have made the changes in GraphQL queries, I have set the limit to fetch limited queries please review.

manasmanohar:
hey @MishraShardendu22, 
- While you have updated the query to limit the number of posts fetched at a time, but aren't we still fetching all the post during build time and sending it our client here? as mentioned earlier a better approach would be, to load data incrementally - fetching only a subset of posts from the API initially, then requesting more as needed upon user clicking load more.
- Load more button is preferred over numbered. 
- Kindly update your pr according to the new pr template. 
- Ensure responsiveness when adding new ui elements (the current numbered pagination ui is not responsive)
image

MishraShardendu22:
Ok sir I'll work on this by today and generate a PR, addressing the issues.

MishraShardendu22:
@amaan-bhati bhaiya can u check and tell what and how I can fix the PR so it can be merged, I was really close to achieving it just some guidance would be nice.

MishraShardendu22:
> Hey @MishraShardendu22 Can you please take a look at the conflict before we move forward with the final review properly in the local?

Sure bhaiya I'll look into this and complete it but next week, thanks !!

PR #121, opened 2025-06-02
Fix minor ui break in mobile devices for table

yaten2302:
## Description

This PR fixes a breaking UI in the mobile device. If the table component is too long, then it takes a lot of extra space.

### Changes

Fixed the CSS for mobile so that the UI doesn't stretch itself, rather it wraps itself and the user can easily scroll through.

### DIFF
--- a/components/post-body.tsx
+++ b/components/post-body.tsx
@@ -280,7 +280,7 @@ export default function PostBody({
       </div>
       <div className={`w-full p-4 ${isList ? "ml-10" : ""}  md:w-4/5 lg:w-3/5`} id="post-body-check">
         {slug === "how-to-compare-two-json-files" && <JsonDiffViewer />}
-        <div className="prose lg:prose-xl">{renderCodeBlocks()}</div>
+        <div className="prose lg:prose-xl prose lg:prose-xl post-content-wrapper">{renderCodeBlocks()}</div>
         <hr className="border-gray-300 mt-10 mb-20" />
         <div>
 
--- a/styles/index.css
+++ b/styles/index.css
@@ -223,3 +223,22 @@ td{
   border: solid 1px black;
   padding: 5px;
 }
+
+.post-content-wrapper table {
+  width: 100%;
+  max-width: 100%;
+  overflow-x: auto;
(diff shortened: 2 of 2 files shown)


manasmanohar:
Hey @yaten2302, please update the PR description to follow the template. Thanks!

styles/index.css
```diff
@@ -222,4 +222,23 @@ pre {
...
+  max-width: 100%;
+  overflow-x: auto;
+  display: block;
+  box-sizing: border-box;
+}
+
+@media (max-width: 1024px) {
+  .post-content-wrapper table {
+    width: 100%;
+    max-width: 100%;
```
manasmanohar, line 238: why are these properties repeated, is it added for any edge case that you observed?

components/post-body.tsx
```diff
@@ -280,7 +280,7 @@ export default function PostBody({
       </div>
       <div className={`w-full p-4 ${isList ? "ml-10" : ""}  md:w-4/5 lg:w-3/5`} id="post-body-check">
         {slug === "how-to-compare-two-json-files" && <JsonDiffViewer />}
-        <div className="prose lg:prose-xl">{renderCodeBlocks()}</div>
+        <div className="prose lg:prose-xl prose lg:prose-xl post-content-wrapper">{renderCodeBlocks()}</div>
```
manasmanohar, line 283: why are these classes added twice?

manasmanohar (changes requested):
also you can remove the uneccesary console logs.
`Here are the heading:`

styles/index.css
yaten2302: No edge case, in here, removing this!

components/post-body.tsx
yaten2302: By mistake added it twice, removing it 👍

manasmanohar (approved):
LGTM!
please look into these points during the table component revamp.
- table cell text alignment https://ux.stackexchange.com/questions/24066/what-is-the-best-practice-for-data-table-cell-content-alignment
- table header text wrapping behavior

PR #174, opened 2025-09-08
Fix marketplace issue 

Bhup-GitHUB:
### Problem
The Keploy blog website was displaying an outdated VS Code extension download count of **540K** instead of the current **696K+** from the marketplace.

### Root Cause
Next.js uses Server-Side Rendering (SSR), which means components render on the server first where browser APIs like `fetch` don't work. The `useEffect` hook that fetches the real-time count only runs on the client-side, but in production, the component was rendering with hardcoded fallback values.

### Solution
Instead of hardcoding values, I implemented a dynamic solution that:
- Fetches real-time data from VS Code Marketplace API
- Uses intelligent caching to reduce API calls  
- Provides multiple fallback mechanisms for reliability
- Works properly with Next.js SSR

### Changes Made
-  Removed hardcoded fallback values for truly dynamic updates
-  Added client-side caching (5min) to reduce API calls
-  Created server-side API route with caching (10min) for better performance
-  Implemented dual data sources: internal API + direct VS Code API
-  Added loading states and graceful error handling
-  Multiple fallback layers ensure count is always displayed

### Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Performance improvement (optimization or efficiency enhancements)

testing
image

Build
image


hooks/useVSCodeInstalls.tsx
```diff
@@ -1,44 +1,101 @@
 import { useEffect, useState } from "react";
 
-export function useVSCodeInstalls(initialInstalls = "540K") {
```
amaan-bhati, line 3: the initial value 540k was written intentionally as the initial installs count, can you please take care of it here?

amaan-bhati (commented):
Hey @Bhup-GitHUB Thanks for raising this pr, please address the comment.

hooks/useVSCodeInstalls.tsx
Bhup-GitHUB: i ll change it to new count number

PR #258, opened 2026-01-04
fix: align navbar width with blog content on scroll

kunalbatra339:
Fixes: #3429

Updated the FloatingNavbar component to use 'max-w-7xl' in the scrolled state. 
This ensures the navbar maintains correct alignment with the main blog content 
instead of shrinking to a narrow pill shape.

Screenshot 2026-01-04 194650

### DIFF
--- a/components/navbar/FloatingNavbar.tsx
+++ b/components/navbar/FloatingNavbar.tsx
@@ -4,8 +4,13 @@ import { useState, useEffect } from "react";
 import { useRouter } from "next/router";
 import FloatingNavbarClient from "./FloatingNavbarClient";
 
+// --- Configuration Constants ---
+const SCROLL_THRESHOLD_START = 50;   // When to start the "pill" shape
+const SCROLL_THRESHOLD_DEEP = 750;   // When to shrink to content width (Hero height)
+
+// --- Styles ---
 const glassNavBase =
-  "relative overflow-visible rounded-full transition-all duration-300 backdrop-blur-2xl";
+  "relative overflow-visible rounded-full transition-all duration-500 ease-[cubic-bezier(0.4,0,0.2,1)] backdrop-blur-2xl"; // Advanced bezier curve for smoother feeling
 const glassNavDefault =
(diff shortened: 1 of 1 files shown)


amaan-bhati (changes requested):
Hey @kunalbatra339 Thanks for raising this pr, the idea of simply increasing the width of the navbar is good, it might help align  the navbar width to the the hero section like this:
<br/>

Image

<br/>

But in cases when the navbar shrinks, it needs to align to the other sections as well, since the navbar remains entact all the time, the ideais to create a good user experience so that they dont need to move their eyes much, hence, when we get to reach these sectios, the navbar needs to shrink to match their width, in this case we can start with the making the navbar have the same width as the hero section as it is right now in case of shrinking, and when we scroll down to other sections, it's width decreses to match their width for an attention to fetail, i have attached the reference images for what i am talking about here for the better understanding of the issue:
<br/>

Image

<br/>

Image
<br/>

amaan-bhati (changes requested):
Hey @kunalbatra339 thanks for following up on the review, i tried it again on the local and have a few suggestions:
- the inital state of the navbar looks good and should stay the same till we get past the hero seciton since the alignment of the navbar matches the hero section in the initial state. image for reference:

Image
<br/>
- The alignment of the navbar when shrinked, instead of getting reduced, it increases and does not matches the width of the other sections and there is a clear gap between the navbar and the sections while scrolling, this gap needs to get removed and the navbar should be aligned in a way that it matches the length of the other sections after scrolling past the hero section, i have attachd the image for reference and better understanding of the issue:

Image

amaan-bhati:
Thank you for the continued work and the thoughtful refinements through multiple review cycles, Kunal! The rAF scroll optimization and the cleaner two-state approach using max-w-7xl for both states are real improvements in response to our feedback. We are closing this PR at this time because the final review feedback we shared - measuring the hero height dynamically instead of hardcoding SCROLL_THRESHOLD = 760, and providing screenshots at multiple viewports - has been open for over three weeks with no updates from the branch. If you want to implement the dynamic height measurement and reopen, we would be happy to take another look! Reach out on Slack at https://keploy.slack.com/join/shared_invite/zt-357qqm9b5-PbZRVu3Yt2rJIa6ofrwWNg if you need guidance.

PR #354, opened 2026-03-31
fix: add missing Script import in _document.tsx

dhananjay6561:
## fix: add missing Script import in `_document.tsx`

**Problem:** Build was failing with `'Script' is not defined` because `next/script` was used in `_document.tsx` without being imported.

**Fix:** Added `import Script from 'next/script'` to resolve the compile error.

**Successful Build:**

Screenshot 2026-03-31 at 5 56 03 PM


amaan-bhati (approved):
LGTM!

Check:

Image

PR #401, opened 2026-07-10
feat: replace sidebar video ad with static image banners

dhananjay6561:
## What & why

Replaces the autoplaying video ad in the blog sidebar with lightweight static **WebP image banners**. Cleaner, image-first promo that loads faster than streaming video.

## Changes
- Swapped the `<video>` element for a `next/image` banner (one of three color variants — cream / dark / orange — shown at random per page load).
- Stripped the banner down to just the **image + a single "Try for Free" CTA**; removed the title, description, "Book a Demo", and "Read the docs" links.
- Added three optimized WebP assets (~17–25 KB each, 720×448).

## Notes
- Images serve under `/blog/images/` (basePath `/blog`) — verified all three return 200.
- `tsc` and `next lint` pass clean.

Screenshot 2026-07-10 at 11 42 33 AM


components/BlogSidebar.tsx
```diff
@@ -102,48 +102,27 @@ function SidebarShare() {
...
-    typeof window !== 'undefined' &&
-    window.matchMedia('(prefers-reduced-motion: reduce)').matches;
-
-  if (!ad) {
+  if (!src) {
     return (
       <div
         className="rounded-2xl bg-white border border-gray-200"
-        style={{ maxWidth: 320, margin: '0 auto', minHeight: 360, boxShadow: '0 4px 24px rgba(0,0,0,0.08)' }}
+        style={{ maxWidth: 320, margin: '0 auto', minHeight: 260, boxShadow: '0 4px 24px rgba(0,0,0,0.08)' }}
```
amaan-bhati: minHeight is 260 here but the actual content ends up taller than that. The image alone is around 199px at this width, plus the padding and the button below it, so it's closer to 285-290px total. Might cause a small jump when the real content swaps in, worth checking in a browser.

components/BlogSidebar.tsx
```diff
@@ -158,93 +137,39 @@ function SidebarAdBanner() {
...
+      <Link
+        href={CTA_HREF}
+        target="_blank"
+        rel="noopener noreferrer"
+        aria-label="Try Keploy for free"
+        style={{ display: 'block', width: '100%' }}
+      >
+        <Image
+          src={src}
+          alt="Try Keploy for free"
```
amaan-bhati: The Link wrapping this already has aria-label="Try Keploy for free" and this alt says the same thing, so screen readers will probably announce it twice for the one link. Since the image is decorative here I'd set alt="" and let the aria-label on the Link carry it.

amaan-bhati (commented):
Nice cleanup, this simplifies the ad unit quite a bit and drops the video weight, which should help load times on post pages. Tested it locally across a few posts, the banner rotates fine between the three variants and both click targets go to the signin page like they should. tsc is clean too.

Left two small comments inline. A couple other things that don't map to a specific line:

- public/images/keploy-ad-banner.jpg was the poster image for the old video and I don't see it referenced anywhere else in the repo anymore. Probably fine to delete it in this PR so it's not just sitting there unused.
- Just want to flag that this cuts more than the video, we go from a headline, description and 3 CTAs down to just the image and one CTA. If that's the intended direction that's fine, just calling it out since it's a bigger scope change than the title suggests.

Otherwise this looks good to me.

dhananjay6561:
Closing in favour of #405, which supersedes this — it's the updated take on the same sidebar-ad rework (static image variants + click tracking, review feedback addressed). Consolidating review there to avoid two PRs rewriting the same BlogSidebar block off main.

PR #405, opened 2026-07-20
feat: replace sidebar video ad with static image banner variants

dhananjay6561:
## Summary

Replaces the sidebar video ad with **4 static image banner variants** (A/B rotation), adds **click tracking** (Clarity + GA4), serves the artwork from **S3**, and makes the banner **fully fluid with a graceful load** and a text fallback.

## What's included

**Banner variants**
- 4 static image variants; the artwork is a full-card image with the **"Try for Free!"** CTA built in code and overlaid in the empty band.
- CTA is anchored to each variant's band center and sized in container-query (`cqw`) units, so it scales with the card and never overflows.
- Random variant selection preserves the existing A/B rotation.

**Responsive & loading**
- Fully fluid: `width: 100%` plus a single shared `aspect-ratio` (no fixed width/height). Sticky sidebar on wide desktop (≥1440px), stacked below the article on smaller screens.
- Reserves the slot from first paint, shows a skeleton shimmer while the image loads, then fades the artwork + CTA in, so there's no layout shift.
- Lazy-loaded at normal priority, since it's a secondary element and shouldn't compete with article content for bandwidth.
- If the image fails or is blocked, it falls back to a styled CTA card so the slot is never blank.

**Click tracking (A/B signal)**
- Fires on the CTA `<Link>` only, so it counts real clickthroughs rather than stray clicks on non-navigating artwork.
- **Clarity** custom tag: `clarity("set", "banner_clicked", <bannerId>)`.
- **GA4** event: `gtag("event", "banner_click", { banner_id })` for exact per-banner counts.
- Both go through a shared readiness/retry helper, so a fast click before Clarity/GA finish loading still lands.

**Assets**
- Banners are served from the `keploy-devrel` S3 bucket; the local `public/images/keploy-ad-banner-*.webp` copies were removed.
- All four variants are now 3x exports (939×1239); banner_3 was re-exported to match after an initial 1x upload.

## Follow-up (not blocking)
- **GA4:** register `banner_id` as an Event-scoped custom dimension so `banner_click` can be broken down per banner in reports.

## Notes
- Supersedes #401 (same `BlogSidebar` rework), which has been closed.

Screenshot 2026-07-20 at 11 20 35 AM


components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
+// (button excluded from the artwork); the "Try for Free!" button is built in
+// code and overlaid in the empty bottom band, with per-variant button colors.
+// The artwork PNGs carry ~8.3% transparent drop-shadow padding, so button
+// positions are measured against the FULL image. btnCenter is the vertical
+// center of each card's empty band (as % from the top); the button is anchored
+// there via translateY(-50%) so it stays centered regardless of its height.
+const AD_BANNERS = [
+  { id: "banner_1", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-1.webp", btnBg: "#ffffff", btnText: "#ED5D0F", btnCenter: "77.8%" },
+  { id: "banner_2", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-2.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "81.6%" },
+  { id: "banner_3", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-3.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "82.0%" },
```
amaan-bhati: Pulled all 4 images from the S3 bucket to check. banner_1, banner_2 and banner_4 are all 939x1239px, but banner_3 is only 313x413px, exactly 1/3 the resolution (939/3=313, 1239/3=413). Looks like it got exported at 1x instead of 3x like the others. It'll render noticeably softer than the other three on any retina/high-DPI screen since the card only has about 1.2x pixel headroom instead of 3.6x. Worth re-exporting banner_3 at the same resolution before this gets real traffic.

components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
+declare global {
+  interface Window {
+    clarity?: (...args: unknown[]) => void;
+    gtag?: (...args: unknown[]) => void;
+  }
+}
+
+function trackBannerClick(bannerId: string) {
+  // Visible in DevTools console so you can confirm clicks are detected.
+  console.log("[ad-banner] click detected → banner_clicked:", bannerId);
```
amaan-bhati: This fires on every click from every visitor since it's not gated behind a dev check. Reads like a debugging aid that was meant to come out before merging (comment even says it's there so "you" can confirm clicks are detected). Would drop it, or wrap it in a NODE_ENV check if it's worth keeping around.

components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
+
+  // GA4 event — gtag buffers into dataLayer, so no readiness guard needed.
+  window.gtag?.("event", "banner_click", { banner_id: bannerId });
```
amaan-bhati: Checked layout.tsx, gtag loads there with strategy="afterInteractive", same async timing class as Clarity's lazyOnload. Clarity gets a full polling retry loop a few lines down in case it's not ready, but this just does window.gtag?.(...) and moves on. If someone clicks the CTA quickly after the page loads, before the afterInteractive script has run, this silently no-ops and that GA4 event is just gone with no retry. Worth giving it the same treatment Clarity got if the point of this PR is reliable A/B data.

components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
-    );
-  }
-
+  // Banner is picked client-side (avoids a hydration mismatch): reserve the
+  // slot, show a skeleton until the image decodes, then fade the artwork in.
   return (
     <div
-      className="rounded-2xl bg-white border border-gray-200 flex flex-col overflow-hidden"
+      data-banner-id={banner?.id}
+      onClick={() => { if (banner) trackBannerClick(banner.id); }}
```
amaan-bhati: This div wraps both the image and the CTA link, so clicking anywhere in the card fires trackBannerClick, including clicks on plain artwork that isn't a link and doesn't navigate anywhere. That means banner_click gets logged for clicks that don't actually convert, which will skew the A/B numbers. Would move this onto the Link itself further down so it only fires for clicks that actually go through to signin.

components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
+              position: 'absolute',
+              inset: 0,
+              width: '100%',
+              height: '100%',
+              objectFit: 'cover',
+              display: 'block',
+              opacity: loaded ? 1 : 0,
+              transition: 'opacity 0.3s ease',
+            }}
+            loading="eager"
```
amaan-bhati: The 3 non-broken images are 170-187KB each (the sibling PR's approach used 17-25KB images for comparison). Marking a sidebar ad eager plus high priority means the browser competes this against the actual article content for bandwidth on every post page view. Seems backwards for a secondary element, and works against the whole point of dropping the video for load time in the first place. Would lazy-load this at normal priority instead.

components/BlogSidebar.tsx
```diff
@@ -103,159 +103,158 @@ function SidebarShare() {
...
   );
 }
 
 /* ── Composed Sidebar ── */
 export default function BlogSidebar() {
   return (
-    <div className="w-full max-w-[260px] flex flex-col gap-5">
+    <>
+      {/* Warm the S3 origin (no crossOrigin — matches the plain <img> fetch). */}
+      <Head>
```
amaan-bhati: BlogSidebar is dynamically imported with ssr: false in post-body.tsx, so this whole component, including this preconnect hint and the image that needs it, only renders client side in the same pass. Preconnect's value is getting a head start before the actual request happens, here there's basically no head start since the img fetch kicks off right alongside it. Don't think this is doing much as written, it would need to live somewhere that renders earlier like _document.tsx to actually help.

amaan-bhati (commented):
Went through this one locally too, checked out the branch, ran tsc and next lint (both clean), and pulled all 4 banner images straight from the S3 bucket to check dimensions against the aspect ratio and button math in the code. Aspect ratio and button positions check out against 3 of the 4 real assets. Left inline comments on a few things worth fixing before this goes out.

One thing that isn't tied to a line: this PR and #401 both rewrite the exact same AD_ITEMS/SidebarAdBanner block in components/BlogSidebar.tsx, off the same main, about 10 days apart. Only one of these can merge cleanly as-is, whichever lands second is going to need a manual rebase against a completely rewritten function. Worth confirming whether this one supersedes #401 so the other can be closed instead of reviewed in parallel.

components/BlogSidebar.tsx
dhananjay6561: yeah took this out completely in 05e4629. ga4 and clarity both record the click already so the log wasnt really adding anything

components/BlogSidebar.tsx
dhananjay6561: good point. pulled the retry logic into a shared helper in a523140 so gtag runs through it too now. if someone clicks before the afterinteractive script loads it just waits and fires once gtag exists, same as clarity

components/BlogSidebar.tsx
dhananjay6561: yep moved the tracking onto the cta link in a523140 so it only counts real clickthroughs, not random clicks on the artwork that dont navigate anywhere

components/BlogSidebar.tsx
dhananjay6561: fair enough, switched it to lazy at normal priority in a523140 so it doesnt fight the article for bandwidth. the file sizes were the bigger issue anyway and banner_3 is sorted now

components/BlogSidebar.tsx
dhananjay6561: true, it wasnt doing anything under ssr false since the component renders client side at the same time as the image. just dropped it in a523140, didnt feel worth moving to _document when were lazy loading now

components/BlogSidebar.tsx
dhananjay6561: yeah youre right, banner_3 got exported at 1x by mistake. theres a proper 3x version now (939x1239), swapped it in at 43a0ec1

components/BlogSidebar.tsx
```diff
@@ -103,151 +103,91 @@ function SidebarShare() {
...
-    primaryCTA: { label: "Get Started", href: "https://app.keploy.io/signin" },
-    secondaryCTA: { label: "Read the docs →", href: "https://keploy.io/docs" },
-  },
+// 4 static banner variants for A/B testing. The artwork is a full-card image
+// (button excluded from the artwork); the "Try for Free!" button is built in
+// code and overlaid in the empty bottom band, with per-variant button colors.
+// The artwork PNGs carry ~8.3% transparent drop-shadow padding, so button
+// positions are measured against the FULL image. btnCenter is the vertical
+// center of each card's empty band (as % from the top); the button is anchored
+// there via translateY(-50%) so it stays centered regardless of its height.
```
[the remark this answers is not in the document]
dhananjay6561: fixed the comment to say webp in a523140, good catch

components/BlogSidebar.tsx
```diff
@@ -103,151 +103,91 @@ function SidebarShare() {
...
     );
   }
 
   return (
     <div
-      className="rounded-2xl bg-white border border-gray-200 flex flex-col overflow-hidden"
       style={{
+        position: 'relative',
+        width: '100%',
         maxWidth: 320,
```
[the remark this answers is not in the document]
dhananjay6561: this ones stale now, the card got reworked in later commits. its a plain img (next/image and the maxwidth 320 are both gone) and the wrapper clips with rounded corners and overflow hidden. sorted as of a523140

components/BlogSidebar.tsx
```diff
@@ -103,151 +103,91 @@ function SidebarShare() {
...
-        >
-          Book a Demo
-        </Link>
+      {/* Full banner artwork (button excluded from artwork) */}
+      <Image
+        src={banner.src}
+        alt="Keploy — 300M+ mocks and 12.8M+ tests generated"
+        width={AD_W}
+        height={AD_H}
+        sizes="320px"
```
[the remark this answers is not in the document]
dhananjay6561: also stale, moved to a plain img so theres no sizes attribute anymore and its lazy loaded at normal priority now (a523140)

amaan-bhati (commented):
Re-reviewed after the latest push. Checked the diff against what I looked at before and confirmed the following are fixed: the debug console.log is gone, gtag and Clarity now go through the same whenReady polling helper so neither silently drops an early click, tracking moved off the card div onto the actual Link so it only fires on real clickthroughs, image loading is back to lazy with no forced fetchPriority, and the preconnect hint got removed instead of left half-useful. tsc and next lint are both still clean. Also checked, #401 is closed now so the conflict between the two PRs is resolved.

Two things left, both inline below.

components/BlogSidebar.tsx
```diff
@@ -103,151 +102,137 @@ function SidebarShare() {
...
+// (button excluded from the artwork); the "Try for Free!" button is built in
+// code and overlaid in the empty bottom band, with per-variant button colors.
+// The .webp artwork carries ~8.3% transparent drop-shadow padding, so button
+// positions are measured against the FULL image. btnCenter is the vertical
+// center of each card's empty band (as % from the top); the button is anchored
+// there via translateY(-50%) so it stays centered regardless of its height.
+const AD_BANNERS = [
+  { id: "banner_1", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-1.webp", btnBg: "#ffffff", btnText: "#ED5D0F", btnCenter: "77.8%" },
+  { id: "banner_2", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-2.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "81.6%" },
+  { id: "banner_3", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/keploy-ad-banner-3.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "82.0%" },
```
amaan-bhati: Re-downloaded all 4 images from S3 just now to check. banner_3 is still 313x413px while banner_1, banner_2 and banner_4 are 939x1239px, same file (26,008 bytes, same Last-Modified timestamp as before). This one didn't get touched, makes sense since it's an S3 asset and wouldn't show up in this diff either way. Still needs a re-export at the same resolution as the other three before this ships, otherwise banner_3 will look visibly softer on retina screens.

components/BlogSidebar.tsx
```diff
@@ -103,151 +102,137 @@ function SidebarShare() {
...
+
+      {banner && (
+        <>
+          {/* Banner artwork (CTA excluded); fills the slot and fades in on load. */}
+          {/* eslint-disable-next-line @next/next/no-img-element */}
+          <img
+            src={banner.src}
+            alt="Keploy — 300M+ mocks and 12.8M+ tests generated"
+            onLoad={() => setLoaded(true)}
+            onError={() => setLoaded(true)}
```
amaan-bhati: Still no fallback here if the image fails to load, onError just sets loaded=true and hides the skeleton. Worth a second look because these S3 filenames (keploy-ad-banner-1.webp through -4.webp) are the kind of pattern ad blockers commonly match on. If that request gets blocked, the visitor gets a broken image icon with the Try for Free button floating over nothing. Doesn't need to be fancy, even just falling back to hiding the button and card if the image errors would be cleaner than the current state.

dhananjay6561: added it back in 0735325. if the image fails or gets blocked it shows a small cta card with the sign up link now instead of just hiding the skeleton, same idea as the old 350 fallback

components/BlogSidebar.tsx
dhananjay6561: sorted in 43a0ec1. banner_3 is the 3x asset now (939x1239, checked it) so it matches the other three. same aspect ratio so nothing else moved

dhananjay6561: sorted in 43a0ec1. banner_3 is the 3x asset now (939x1239, checked it) so it matches the other three. same aspect ratio so nothing else moved

components/BlogSidebar.tsx
dhananjay6561: sorted in 43a0ec1. banner_3 is the 3x asset now (939x1239, checked it) so it matches the other three. same aspect ratio so nothing else moved

components/BlogSidebar.tsx
```diff
@@ -103,151 +102,164 @@ function SidebarShare() {
...
+  // Text fallback so the ad slot is never blank if the artwork fails to load
+  // (content blocker, S3 hiccup, etc.). Keeps a working CTA + click tracking.
+  if (banner && errored) {
     return (
-      <div
-        className="rounded-2xl bg-white border border-gray-200"
-        style={{ maxWidth: 320, margin: '0 auto', minHeight: 360, boxShadow: '0 4px 24px rgba(0,0,0,0.08)' }}
-        aria-hidden="true"
-      />
+      <div className="rounded-2xl p-5 flex flex-col justify-center" style={{ backgroundColor: '#FFF4EE' }}>
```
amaan-bhati: This fallback doesn't reserve the same height as the normal loaded state (no aspectRatio here, unlike the card below). If the image genuinely fails mid-session, the card will visibly shrink and shift the divider/newsletter form below it. Better than a broken image for sure, just flagging that it trades "broken image" for "layout shift on failure." Might be worth wrapping this in the same aspect-ratio box, vertically centered, so there's no shift either way.

components/BlogSidebar.tsx
```diff
@@ -103,151 +102,164 @@ function SidebarShare() {
...
+          Generate test cases and data mocks with one click. Reduce unit test development time by 90%.
+        </p>
+        <Link
+          href={CTA_HREF}
+          target="_blank"
+          rel="noopener noreferrer"
+          data-banner-id={banner.id}
+          onClick={() => trackBannerClick(banner.id)}
+          className="inline-flex items-center gap-1 font-semibold text-sm transition-colors duration-150 hover:opacity-80"
+          style={{ fontFamily: "'DM Sans', sans-serif", color: '#FF6D41' }}
```
amaan-bhati: Small one: this uses #FF6D41, which doesn't match the brand orange used everywhere else in this file (#ED5D0F, used for banner_2/banner_3's button background and originally for the old primary CTA). Probably just want to reuse #ED5D0F here for consistency, doesn't need to be a new color.

amaan-bhati (commented):
Went through the full diff against main this time, not just what changed since my last pass, checked out the branch, ran tsc and next lint (both clean), re-downloaded all 4 banner images from S3 directly, and rendered the component live including simulating a real image load failure to confirm the fallback actually works, not just reads correctly.

Confirmed fixed from earlier rounds: banner_3 is now 939x1239 matching the other three, the debug console.log is gone, gtag and Clarity both go through the same whenReady retry so neither drops an early click, click tracking is scoped to the actual CTA link not the whole card, image loading is lazy with no forced priority, and the fallback text renders correctly with a working tracked link when the image fails (I forced this with a route abort in a real browser, not just read the code).

Two small things left inline. Also, unrelated to this diff but caused by it: public/images/keploy-ad-banner.jpg (1.9MB, the old video's poster image) has zero references anywhere in the repo now. Not part of this PR since it was already sitting in main, but this PR is what makes it dead weight. Worth deleting here or in a fast-follow.

components/BlogSidebar.tsx
dhananjay6561: good catch. wrapped the fallback in the same aspect ratio box and centered it vertically in de7294d so it holds the same height as the loaded card, no shift now even if the image dies mid session

components/BlogSidebar.tsx
dhananjay6561: yep switched it to #ED5D0F in de7294d to match the rest of the file, no reason for a separate orange here

amaan-bhati (commented):
Hey @dhananjay6561 I tried reviewing the pr locally as well, there are a few attention to details we can iterate on in my opinion. 

The ui of the ad banner looks good, it is also changing on refresh which makes our goal to try multiple ad banners successfull as well. The attention to detail here is on the width of the ad banner, the section right above and below it are the share on socials and the newsletter section, both of these sections more in width compared to the ad banner. 

The width of ad banner is lesser than both and is not mathciing the aligment width wise, i think that either we can iterate on increasing the width of the ad banner(to fit the images, we can maintain the same ratio by increasing both height and width as well. I am adding the screenshots for your reference as well:

- The Difference in the width:
  - Desktop:
Image

<br/>

  -  Mobile:

Image

- Skeleton loader for the ads banner also has more width than the actual size of the ad banner, both in terms of the height and the width, (the width of the skeleton laoder for this banner is the same as the other two sections above and below the banner)

Image

amaan-bhati (approved):
LGTM

PR #408, opened 2026-07-28
feat: updated ui for the 5 year cta banner

amaan-bhati:
Rebuilt the inline "5 Years of Keploy" promo CTA banner (`components/InlinePromoCard.tsx`) from scratch against the design team's final Figma file. Placement of the banner within each blog post is unchanged — only the card itself changed, and this replaces the previous gradient/animated-border version entirely rather than iterating on it.

**Desktop**
- Full-width card matching the blog's content column 
- 3 decorative orange ellipse "blobs" (top-right, left-side, bottom) clipped by the card edges, matched to the Figma reference by cross-checking pixel positions against the exported design frames.
- Badge (5-years anniversary ribbon) is absolutely positioned so its `top`/`right` offsets sit outside the card's normal content box, then gets partially clipped by the card's `overflow: hidden` — matching the same clip-content behavior the badge has inside its own Figma frame. This is what lets the card's height be driven purely by the text/button content rather than by the badge's own size.
- Heading wraps to 3 lines per spec: "Celebrate with us / Get **1 Month of FREE** / Keploy Credits" (Lexend Bold). "1 Month of FREE" is one gradient-coloured phrase (`#FF9D2E → #F76B1C`), rest of the heading is plain black - rendered as its own dedicated heading variant (see Mobile below)
- CTA changed from an animated gradient pill to a solid `#F76B1C` pill with a diagonal arrow icon - "Claim 1 Month of Free Credits". Focus ring updated to `#F76B1C` (was a leftover `#F59E0B` from the old palette).
- Dropped the old animated rainbow border, shine-sweep animation, and gradient watercolor wash

**Mobile (≤850px)**
- Centered, stacked layout: badge → heading (2-line variant, since the vertical stack has more relative width to work with) → CTA
- Badge swapped to a separately-cropped asset (`badge-tight.png`) instead of the desktop one - the desktop asset's large transparent canvas padding was creating an oversized gap between badge and heading once stacked vertically. It sits in normal document flow (not absolutely positioned) so its layout doesn't depend on the image's exact intrinsic height.
- Breakpoint widened from 600px to 850px after review caught the 601–850px range (tablets, small laptops) crushing the heading into 6-7 lines against the fixed-width badge. Verified the blog's content column locks to a fixed 732px width from 850px through 1920px, so the desktop layout never gets squeezed above that point.
- Heading at 14px / weight 800 (up from 700) - tested up to 16px, but 15px already wraps to 3 lines on an actual 375px-wide iPhone SE, so 14px is the real ceiling that holds 2 lines on both SE and iPhone 14 widths
- CTA at a compact ~10px/5px-9px padding - kept close to the original Figma spec after a larger version (16px font / 30px padding) was overflowing the card frame on real iPhone SE and iPhone 14 widths
- Desktop and mobile now render as two independent heading elements toggled by CSS (`k5y-heading-desktop` / `k5y-heading-mobile`) instead of hiding a `<br>` via `display: none`, which was inconsistent across renderers and broke if the copy ever changed

**Accessibility**
- Restored a screen-reader-only description ("Celebrating 5 years of Keploy — claim one month of free credits.") linked to the CTA via `aria-describedby`, matching what the previous version had
- All 9 images now carry explicit `width`/`height` attributes (previously missing, a layout-shift risk). The 7 decorative ellipses load `loading="lazy"` + `decoding="async"`; the two badge images (the visual centerpiece of the card) load `eager` instead, since lazy-loading the largest, most prominent image in the component was causing a visible pop-in

**Code health**
- Moved all component CSS out of an inline `<style>` template string into `components/InlinePromoCard.module.css`, matching how the rest of this repo already scopes component styles (`subscribe-newsletter.module.css`, `post-body.module.css`, etc.) - now it's scoped, goes through the normal lint/build pipeline, and doesn't re-inject a global `<style>` tag on every render
- Removed the `"use client"` directive - this repo is Pages Router only (no `app/` directory), so it was inert

**Assets**
- All 9 images (badge, cropped mobile badge, 7 ellipse SVGs) now load from `keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/` - nothing for this component is bundled in `public/` anymore
- Added the `Lexend` Google Font (weights 500/700/800 - only what's actually used) to `pages/_document.tsx`, following the same preload+stylesheet pattern already used there for Baloo 2 / DM Sans

**Known open items** (see PR conversation)
- `config/inline-promos.ts` anchor text change predates this redesign and moves the CTA from the top to the bottom of the `software-testing-basics` post - still needs an explicit call on whether to revert or keep
- Mobile CTA is intentionally compact to avoid the SE/14 overflow described above, but is under the ~44px touch-target guideline - open to a follow-up if that needs addressing before merge
- The CTA button and heading accent colour (`#F76B1C`) measure below WCAG AA contrast on white (2.97:1, needs 4.5:1) - this was raised in review, but the colour is part of the signed-off Figma design, so it's being kept as-is rather than adjusted unilaterally

### Desktop:
Screenshot 2026-07-31 at 9 42 20 PM

### Mobile:
Screenshot 2026-07-31 at 9 44 50 PM

### Have also created a figma design for the same earlier:
- Iterations: https://www.figma.com/design/6Xw3GmXqvWvj52o1wJ45i3/Keploy---New-file--Copy-?node-id=2248-1094&t=m8fQ0MNRN1MswCuJ-1
- Final ui: https://www.figma.com/design/6Xw3GmXqvWvj52o1wJ45i3/Keploy---New-file--Copy-?node-id=2248-1093&m=dev

- Ran `npm run build`, nothing breaks:
Screenshot 2026-07-28 at 3 06 12 PM

### DIFF
--- a/components/InlinePromoCard.tsx
+++ b/components/InlinePromoCard.tsx
@@ -78,7 +78,7 @@ function Keploy5YearsBanner() {
           display: inline-flex;
           align-items: center;
           gap: 6px;
-          padding: 9.6px 18px;
+          padding: 10px 18px;
           background: #f76b1c;
           color: #ffffff;
           border-radius: 9999px;
@@ -96,29 +96,27 @@ function Keploy5YearsBanner() {
         .k5y-cta-btn:focus-visible { outline: 3px solid #f59e0b; outline-offset: 2px; }
         .k5y-cta-icon { width: 18px; height: 18px; flex-shrink: 0; }
 
-        @media (max-width: 600px) {
+        @media (max-width: 850px) {
           .k5y-ellipse-tr, .k5y-ellipse-left, .k5y-ellipse-bottom { display: none; }
           .k5y-ellipse-tl-m { display: block; top: 0; left: 0; width: 54px; height: 96px; }
           .k5y-ellipse-top-m { display: block; top: 0; right: 46px; width: 61px; height: 28px; }
           .k5y-ellipse-right-m { display: block; top: 35px; right: 0; width: 41px; height: 166px; }
           .k5y-ellipse-bottom-m { display: block; left: 23px; bottom: 0; width: 53px; height: 33px; }
 
-          .k5y-inner { text-align: center; padding: 250px 16px 44px; }
+          .k5y-inner { text-align: center; padding: 0 16px 44px; }
           .k5y-badge-img { display: none; }
           .k5y-badge-img-mobile {
             display: block;
-            position: absolute;
-            top: 32px;
-            left: 50%;
+            position: static;
             width: 190px;
             height: auto;
-            transform: translateX(-50%);
+            margin: 32px auto 16px;
           }
 
-          .k5y-heading { font-size: 12.5px; font-weight: 800; line-height: 1.35; margin: 0 0 22px; }
+          .k5y-heading { font-size: 12px; font-weight: 800; line-height: 1.35; margin: 0 0 22px; }
           .k5y-lb2 { display: none; }
 
-          .k5y-cta-btn { padding: 4.8px 8.8px; font-size: 10px; gap: 3px; }
+          .k5y-cta-btn { padding: 5px 9px; font-size: 10px; gap: 3px; }
           .k5y-cta-icon { width: 10px; height: 10px; }
         }
       `}</style>
@@ -126,24 +124,28 @@ function Keploy5YearsBanner() {
       <div className="k5y-card">
         {/* Decorative blobs — desktop */}
         {/* eslint-disable @next/next/no-img-element */}
-        <img className="k5y-ellipse k5y-ellipse-tr" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-right.svg" alt="" aria-hidden="true" />
-        <img className="k5y-ellipse k5y-ellipse-left" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-left-side.svg" alt="" aria-hidden="true" />
-        <img className="k5y-ellipse k5y-ellipse-bottom" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-bottom.svg" alt="" aria-hidden="true" />
+        <img className="k5y-ellipse k5y-ellipse-tr" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-right.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-left" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-left-side.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-bottom" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-bottom.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
         {/* Decorative blobs — mobile */}
-        <img className="k5y-ellipse k5y-ellipse-tl-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-left-mobile.svg" alt="" aria-hidden="true" />
-        <img className="k5y-ellipse k5y-ellipse-top-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-mobile-top.svg" alt="" aria-hidden="true" />
-        <img className="k5y-ellipse k5y-ellipse-right-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-right-mobile.svg" alt="" aria-hidden="true" />
-        <img className="k5y-ellipse k5y-ellipse-bottom-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-bottom-mobile.svg" alt="" aria-hidden="true" />
+        <img className="k5y-ellipse k5y-ellipse-tl-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-left-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-top-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-mobile-top.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-right-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-right-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-bottom-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-bottom-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
 
         <img
           className="k5y-badge-img"
           src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/badge.png"
           alt="Keploy 5 years anniversary badge"
+          loading="lazy"
+          decoding="async"
         />
         <img
           className="k5y-badge-img-mobile"
           src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/badge-tight.png"
           alt="Keploy 5 years anniversary badge"
+          loading="lazy"
+          decoding="async"
         />
 
         <div className="k5y-inner">
--- a/components/InlinePromoCard.tsx
+++ b/components/InlinePromoCard.tsx
@@ -15,50 +15,52 @@ function Keploy5YearsBanner() {
           0%        { transform: translateX(-120%) skewX(-12deg); }
           65%, 100% { transform: translateX(600%) skewX(-12deg); }
         }
-        .k5y-body { display: flex; align-items: center; gap: 20px; }
-        .k5y-card { padding: 22px 28px; }
+        /* Desktop: bigger centered card, single vertical column (same shape as mobile, scaled up) */
+        .k5y-body { display: flex; flex-direction: column; align-items: center; text-align: center; gap: 20px; }
+        .k5y-card { max-width: 460px; margin: 0 auto; padding: 44px 40px; border-radius: 24px; }
         .k5y-badge-wrap {
           position: relative;
-          width: 84px;
-          height: 85px;
+          width: 190px;
+          height: 196px;
           flex-shrink: 0;
           overflow: hidden;
-          transform: translateY(-7px);
+          margin: 0 auto;
         }
         .k5y-badge-img {
           position: absolute;
           top: 50%;
           left: 50%;
-          height: 124px;
+          height: 282px;
           width: auto;
           max-width: none;
           transform: translate(-50%, -50%);
         }
-        .k5y-divider { display: block; }
+        .k5y-content { text-align: center; }
         .k5y-heading {
           color: #171412;
-          font-size: 16px;
+          font-size: 24px;
           font-weight: 700;
-          margin: 0 0 2px;
+          margin: 0 0 10px;
           line-height: 1.35;
         }
         .k5y-desc {
           color: #92400e;
-          font-size: 13.5px;
+          font-size: 15.5px;
           margin: 0;
-          line-height: 1.65;
+          line-height: 1.6;
         }
         .k5y-cta-btn {
           flex-shrink: 0;
           display: inline-flex;
           align-items: center;
           justify-content: center;
-          padding: 12px 12px;
+          margin-top: 24px;
+          padding: 14px 28px;
           background: linear-gradient(180deg, #ff9d2e 0%, #ff6b00 100%);
           color: white;
           border: none;
-          border-radius: 10px;
-          font-size: 13.5px;
+          border-radius: 12px;
+          font-size: 15px;
           font-weight: 700;
           text-decoration: none;
           box-shadow: 0 5px 14px rgba(255,107,0,0.14);
@@ -80,18 +82,18 @@ function Keploy5YearsBanner() {
           outline: 3px solid #f59e0b;
           outline-offset: 2px;
         }
+        /* Mobile: same vertical shape, scaled back down, full-width thumb-friendly button */
         @media (max-width: 600px) {
-          .k5y-body { flex-direction: column; align-items: center; text-align: center; gap: 14px; }
-          .k5y-badge-wrap { width: 150px; height: 155px; transform: none; margin: 0 auto; }
+          .k5y-body { gap: 14px; }
+          .k5y-card { max-width: 100%; padding: 20px 18px; border-radius: 16px; }
+          .k5y-badge-wrap { width: 150px; height: 155px; }
           .k5y-badge-img { height: 224px; }
-          .k5y-divider { display: none; }
-          .k5y-content { text-align: center; }
           .k5y-heading { font-size: 21px; margin: 0 0 8px; }
           .k5y-desc { font-size: 14px; line-height: 1.5; }
-          .k5y-card { padding: 20px 18px; }
           .k5y-cta-btn {
             width: 100%;
             white-space: normal;
+            margin-top: 14px;
             padding: 15px 20px;
             border-radius: 16px;
             font-size: 15.5px;
@@ -122,7 +124,7 @@ function Keploy5YearsBanner() {
             height: "100%",
             background:
               "linear-gradient(105deg, transparent 30%, rgba(251,191,36,0.10) 50%, transparent 70%)",
-            animation: "k5y-sweep 4.5s ease-in-out infinite",
+            animation: "k5y-sweep 3s ease-in-out infinite",
             pointerEvents: "none",
           }}
         />
@@ -137,40 +139,25 @@ function Keploy5YearsBanner() {
           />
         </div>
 
-        {/* Vertical divider */}
-        <div
-          className="k5y-divider"
-          style={{
-            width: 1.5,
-            alignSelf: "stretch",
-            flexShrink: 0,
-            opacity: 0.35,
-            background:
-              "linear-gradient(to bottom, transparent, #fdba74 15%, #f97316 50%, #fdba74 85%, transparent)",
-          }}
-        />
-
-        {/* Text */}
-        <div className="k5y-content" style={{ flex: 1, minWidth: 0 }}>
+        {/* Text + CTA, grouped together as one section */}
+        <div className="k5y-content" style={{ minWidth: 0 }}>
           <p className="k5y-heading">Keploy Turns 5 This Month!</p>
           <p id={`${bannerId}-desc`} className="k5y-desc">
             To celebrate our 5 years, we&apos;re giving away
             <br />
             one month of Keploy credits for free.
           </p>
+          <a
+            href="https://keploy.io/credits-form"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="k5y-cta-btn"
+            aria-label="Claim 1 month of free Keploy credits"
+            aria-describedby={`${bannerId}-desc`}
+          >
+            Claim 1 Month of Free Credits
+          </a>
         </div>
-
-        {/* CTA */}
-        <a
-          href="https://keploy.io/credits-form"
-          target="_blank"
-          rel="noopener noreferrer"
-          className="k5y-cta-btn"
-          aria-label="Claim 1 month of free Keploy credits"
-          aria-describedby={`${bannerId}-desc`}
-        >
-          Claim 1 Month of Free Credits
-        </a>
       </div>
     </div>
   );
--- a/components/InlinePromoCard.tsx
+++ b/components/InlinePromoCard.tsx
@@ -15,24 +15,39 @@ function Keploy5YearsBanner() {
           0%        { transform: translateX(-120%) skewX(-12deg); }
           65%, 100% { transform: translateX(600%) skewX(-12deg); }
         }
-        .k5y-body { display: flex; align-items: center; gap: 26px; }
+        .k5y-body { display: flex; align-items: center; gap: 20px; }
+        .k5y-card { padding: 22px 28px; }
         .k5y-badge-wrap {
           position: relative;
-          width: 88px;
-          height: 90px;
+          width: 84px;
+          height: 85px;
           flex-shrink: 0;
           overflow: hidden;
+          transform: translateY(-7px);
         }
         .k5y-badge-img {
           position: absolute;
           top: 50%;
           left: 50%;
-          height: 130px;
+          height: 124px;
           width: auto;
           max-width: none;
           transform: translate(-50%, -50%);
         }
         .k5y-divider { display: block; }
+        .k5y-heading {
+          color: #171412;
+          font-size: 16px;
+          font-weight: 700;
+          margin: 0 0 2px;
+          line-height: 1.35;
+        }
+        .k5y-desc {
+          color: #92400e;
+          font-size: 13.5px;
+          margin: 0;
+          line-height: 1.65;
+        }
         .k5y-cta-btn {
           flex-shrink: 0;
           display: inline-flex;
@@ -46,7 +61,7 @@ function Keploy5YearsBanner() {
           font-size: 13.5px;
           font-weight: 700;
           text-decoration: none;
-          box-shadow: 0 5px 14px rgba(255,107,0,0.16);
+          box-shadow: 0 5px 14px rgba(255,107,0,0.14);
           letter-spacing: 0.01em;
           cursor: pointer;
           white-space: nowrap;
@@ -54,7 +69,7 @@ function Keploy5YearsBanner() {
           font-family: inherit;
         }
         .k5y-cta-btn:hover {
-          box-shadow: 0 6px 16px rgba(255,107,0,0.20);
+          box-shadow: 0 6px 16px rgba(255,107,0,0.18);
           transform: translateY(-1px);
           filter: brightness(1.05);
         }
@@ -66,24 +81,33 @@ function Keploy5YearsBanner() {
           outline-offset: 2px;
         }
         @media (max-width: 600px) {
-          .k5y-body { flex-direction: column; align-items: flex-start; gap: 16px; }
-          .k5y-badge-wrap { width: 70px; height: 72px; }
-          .k5y-badge-img { height: 104px; }
+          .k5y-body { flex-direction: column; align-items: center; text-align: center; gap: 14px; }
+          .k5y-badge-wrap { width: 150px; height: 155px; transform: none; margin: 0 auto; }
+          .k5y-badge-img { height: 224px; }
           .k5y-divider { display: none; }
-          .k5y-cta-btn { width: 100%; text-align: center; white-space: normal; }
+          .k5y-content { text-align: center; }
+          .k5y-heading { font-size: 21px; margin: 0 0 8px; }
+          .k5y-desc { font-size: 14px; line-height: 1.5; }
+          .k5y-card { padding: 20px 18px; }
+          .k5y-cta-btn {
+            width: 100%;
+            white-space: normal;
+            padding: 15px 20px;
+            border-radius: 16px;
+            font-size: 15.5px;
+          }
         }
       `}</style>
 
       {/* Card body */}
       <div
-        className="k5y-body"
+        className="k5y-body k5y-card"
         style={{
           background:
-            "radial-gradient(65% 110% at 0% 45%, rgba(255,140,32,0.12) 0%, rgba(255,157,46,0.06) 24%, rgba(255,255,255,0) 46%), #ffffff",
+            "radial-gradient(85% 140% at 0% 45%, rgba(255,140,32,0.10) 0%, rgba(255,150,40,0.06) 20%, rgba(255,160,50,0.025) 38%, rgba(255,255,255,0) 64%), #ffffff",
           borderRadius: 16,
           border: "1px solid #e5e7eb",
           boxShadow: "0 4px 24px rgba(0,0,0,0.08)",
-          padding: "22px 28px",
           position: "relative",
           overflow: "hidden",
         }}
@@ -120,35 +144,19 @@ function Keploy5YearsBanner() {
             width: 1.5,
             alignSelf: "stretch",
             flexShrink: 0,
-            opacity: 0.7,
+            opacity: 0.35,
             background:
               "linear-gradient(to bottom, transparent, #fdba74 15%, #f97316 50%, #fdba74 85%, transparent)",
           }}
         />
 
         {/* Text */}
-        <div style={{ flex: 1, minWidth: 0 }}>
-          <p
-            style={{
-              color: "#171412",
-              fontSize: 16,
-              fontWeight: 700,
-              margin: "0 0 6px",
-              lineHeight: 1.35,
-            }}
-          >
-            Keploy turned 5 this month!
-          </p>
-          <p
-            id={`${bannerId}-desc`}
-            style={{
-              color: "#92400e",
-              fontSize: 13.5,
-              margin: 0,
-              lineHeight: 1.65,
-            }}
-          >
-            To celebrate our 5 years, we are giving away one month of Keploy credits for free!
+        <div className="k5y-content" style={{ flex: 1, minWidth: 0 }}>
+          <p className="k5y-heading">Keploy Turns 5 This Month!</p>
+          <p id={`${bannerId}-desc`} className="k5y-desc">
+            To celebrate our 5 years, we&apos;re giving away
+            <br />
+            one month of Keploy credits for free.
           </p>
         </div>
 
--- a/components/InlinePromoCard.tsx
+++ b/components/InlinePromoCard.tsx
@@ -17,42 +17,38 @@ function Keploy5YearsBanner() {
           100% { background-position: 0% 50%; }
         }
         @keyframes k5y-glow {
-          0%, 100% { box-shadow: 0 2px 16px rgba(0,0,0,0.06), 0 0 0 0 rgba(251,176,45,0); }
-          50%       { box-shadow: 0 2px 20px rgba(0,0,0,0.07), 0 0 20px 4px rgba(251,176,45,0.16); }
+          0%, 100% { box-shadow: 0 2px 16px rgba(0,0,0,0.05), 0 0 0 0 rgba(251,146,60,0); }
+          50%       { box-shadow: 0 2px 20px rgba(0,0,0,0.06), 0 0 18px 3px rgba(251,146,60,0.14); }
         }
         @keyframes k5y-sweep {
           0%        { transform: translateX(-120%) skewX(-12deg); }
           65%, 100% { transform: translateX(600%) skewX(-12deg); }
         }
-        @keyframes k5y-sparkle {
-          0%, 100% { opacity: 0.55; transform: scale(1) rotate(0deg); }
-          50%       { opacity: 1;    transform: scale(1.25) rotate(18deg); }
-        }
-        .k5y-body { display: flex; align-items: center; gap: 24px; }
-        .k5y-badge { display: flex; flex-direction: column; align-items: center; gap: 7px; }
+        .k5y-body { display: flex; align-items: center; gap: 26px; }
+        .k5y-badge-img { display: block; height: 76px; width: auto; flex-shrink: 0; }
         .k5y-divider { display: block; }
         .k5y-cta-btn {
           flex-shrink: 0;
           display: inline-flex;
           align-items: center;
           justify-content: center;
-          padding: 10px 20px;
-          background: linear-gradient(90deg, #f59e0b, #f97316);
+          padding: 12px 22px;
+          background: linear-gradient(135deg, #fb923c, #f97316);
           color: white;
           border: none;
           border-radius: 10px;
-          font-size: 13px;
+          font-size: 13.5px;
           font-weight: 700;
           text-decoration: none;
-          box-shadow: 0 3px 14px rgba(249,115,22,0.32);
-          letter-spacing: 0.02em;
+          box-shadow: 0 4px 12px rgba(249,115,22,0.28);
+          letter-spacing: 0.01em;
           cursor: pointer;
           white-space: nowrap;
-          transition: opacity 0.15s ease, transform 0.15s ease;
+          transition: box-shadow 0.15s ease, transform 0.15s ease;
           font-family: inherit;
         }
         .k5y-cta-btn:hover {
-          opacity: 0.9;
+          box-shadow: 0 6px 16px rgba(249,115,22,0.36);
           transform: translateY(-1px);
         }
         .k5y-cta-btn:focus-visible {
@@ -60,8 +56,8 @@ function Keploy5YearsBanner() {
           outline-offset: 2px;
         }
         @media (max-width: 600px) {
-          .k5y-body { flex-direction: column; align-items: flex-start; gap: 14px; }
-          .k5y-badge { flex-direction: row; align-items: center; gap: 8px; }
+          .k5y-body { flex-direction: column; align-items: flex-start; gap: 16px; }
+          .k5y-badge-img { height: 64px; }
           .k5y-divider { display: none; }
           .k5y-cta-btn { width: 100%; text-align: center; white-space: normal; }
         }
@@ -82,7 +78,9 @@ function Keploy5YearsBanner() {
         <div
           className="k5y-body"
           style={{
-            background: "#fffcf7",
+            background:
+              "radial-gradient(55% 100% at 0% 50%, rgba(251,146,60,0.16) 0%, rgba(253,186,116,0.07) 22%, rgba(255,255,255,0) 42%), " +
+              "#ffffff",
             borderRadius: "calc(16px - 1.5px)",
             padding: "22px 28px",
             position: "relative",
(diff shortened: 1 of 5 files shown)


dhananjay6561 (commented):
Strict review. The rewrite is genuinely cleaner in places — dropping `useId`, killing the animated-border/shine/glow machinery, extracting `ArrowIcon`. But there are two things I'd treat as blocking-until-answered and a cluster of fragility/perf issues. Details inline; summary here.

**Blocking / needs a decision**
1. `config/inline-promos.ts` silently **relocates** the CTA. The PR body says *“Placement of the banner within each blog post is unchanged”* — but changing `afterText` is exactly what changes placement. I checked the live post: both the old and new anchor phrases exist, so it won’t 404 — but the old anchor is the **opening** definition and the new one is a **concluding** paragraph, so the banner moves from the top of the article to the bottom. That’s a conversion decision, not a UI tweak, and it’s buried in a “refresh the card” PR.
2. There is exactly **one** breakpoint (600px). Nothing covers ~601–800px (tablets / small laptops), where the desktop rule with 349px right-padding + a fixed 337px badge will crush the text column and/or overlap it.

**Description ↔ code mismatches** (reviewers trust the description)
3. Mobile heading: body says *15px*, code says `12.5px`.
4. Mobile CTA: body says *“16px font / 30px padding”*, code says `10px` font / `4.8px` padding — a ~19px-tall tap target, well under the ~44px touch guideline.

**Fragility / perf**
5. Layout is pinned to specific external-asset pixel dimensions via magic numbers (349px, 250px, sub-pixel 4.8/9.6px).
6. 9 external S3 image requests for *decoration* on every promoted post, no `loading=lazy` / `decoding=async`, no fallback if the bucket URL rotates. The previous version drew all of this with zero requests (inline CSS gradients). Given the org literally just had a badge break from an external-dependency change, this one is worth taking seriously.

Nothing here is unfixable; #1 and #2 just need an actual answer before merge.

config/inline-promos.ts
```diff
@@ -10,7 +10,7 @@ export const inlinePromos: InlinePromoConfig[] = [
   {
     // https://keploy.io/blog/community/software-testing-basics
     blogSlug: "software-testing-basics",
-    afterText: "Software testing is the process of evaluating a software",
+    afterText: "There is no doubt that having a strong foundation",
```
dhananjay6561, line 13: **This is load-bearing and contradicts the PR description.** `post-body.tsx` builds a regex from `afterText` and injects the card after the first block containing it; if nothing matches, `applyPromos` falls through and the post renders **with no promo at all** (no fallback).

I verified against the live `software-testing-basics` post: the new phrase *does* exist, so it won’t drop — **but** the old anchor (“Software testing is the process of evaluating a software”) is the article’s opening definition, while this new one is a concluding paragraph. So this silently moves the CTA from the **top** of the post to the **bottom**. The PR body says “Placement of the banner within each blog post is unchanged” — that’s not what this line does.

Either revert this to keep placement truly unchanged, or call the move out explicitly and get product/DevRel sign-off — pushing the CTA to the end of the article is a conversion decision, not a styling one. Also worth a comment noting the silent-drop failure mode if the CMS copy ever changes.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
+          display: none;
         }
-        @keyframes k5y-sweep {
-          0%        { transform: translateX(-120%) skewX(-12deg); }
-          65%, 100% { transform: translateX(600%) skewX(-12deg); }
+
+        .k5y-inner {
+          position: relative;
+          z-index: 1;
+          padding: 44px 349px 44px 56px;
```
dhananjay6561: `349px` right padding is a magic number hard-reserving space for the absolutely-positioned 337px badge, and there is only **one** breakpoint (600px). For every viewport between 601px and ~800px (tablets, small laptops, split-screen) this desktop rule applies: 349 + 56 = **405px** of horizontal padding. If the blog content column is ~680px, the heading is left ~275px while a 337px badge sits on top of it — guaranteed overlap/crush in that range. Needs an intermediate breakpoint, or a layout that isn’t pinned to a fixed badge width (e.g. badge sized in %/max-width, padding derived from it).

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
+        }
+        .k5y-badge-img {
+          position: absolute;
+          z-index: 1;
+          top: -21px;
```
dhananjay6561: The badge is offset to `top: -21px; right: -12px` to “bleed past the card’s top/right edges” (per the PR body) — but `.k5y-card` has `overflow: hidden` (line 32), which clips exactly those bleed regions. So either the intended bleed is being silently cropped, or the description is wrong. Please confirm which is intended. Separately, `337×337` fixed px means on any content column narrower than ~349px the badge dominates the card.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
-          .k5y-badge { flex-direction: row; align-items: center; gap: 8px; }
-          .k5y-divider { display: none; }
-          .k5y-cta-btn { width: 100%; text-align: center; white-space: normal; }
+          .k5y-ellipse-tr, .k5y-ellipse-left, .k5y-ellipse-bottom { display: none; }
+          .k5y-ellipse-tl-m { display: block; top: 0; left: 0; width: 54px; height: 96px; }
+          .k5y-ellipse-top-m { display: block; top: 0; right: 46px; width: 61px; height: 28px; }
+          .k5y-ellipse-right-m { display: block; top: 35px; right: 0; width: 41px; height: 166px; }
+          .k5y-ellipse-bottom-m { display: block; left: 23px; bottom: 0; width: 53px; height: 33px; }
+
+          .k5y-inner { text-align: center; padding: 250px 16px 44px; }
```
dhananjay6561: `250px` top padding is a magic number reserving vertical room for the absolutely-positioned mobile badge — which is sized `width: 190px; height: auto`. Because the height is intrinsic to `badge-tight.png`, its real rendered height is unknown at CSS-authoring time, so 250px is a guess tuned to today’s asset. Re-crop or swap that PNG and this silently over- or under-reserves → gap or overlap with the heading. Pin the badge to a known aspect ratio (or give the img explicit width+height) instead of hand-tuning the parent’s padding.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
+            display: block;
+            position: absolute;
+            top: 32px;
+            left: 50%;
+            width: 190px;
+            height: auto;
+            transform: translateX(-50%);
+          }
+
+          .k5y-heading { font-size: 12.5px; font-weight: 800; line-height: 1.35; margin: 0 0 22px; }
```
dhananjay6561: Code sets the mobile heading to `12.5px`, but the PR description says “Heading bumped to **15px** / weight 800.” 2.5px is a visible difference on the primary message — reconcile description vs code.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
+          .k5y-lb2 { display: none; }
+
+          .k5y-cta-btn { padding: 4.8px 8.8px; font-size: 10px; gap: 3px; }
```
dhananjay6561: Two problems on one line:
1. **Tap target / legibility:** `font-size: 10px` with `4.8px` vertical padding is a ~19px-tall button — far below the ~44px minimum for touch targets, and 10px is barely readable on mobile. This is the primary CTA of the banner.
2. **Contradicts the description:** the PR body says the mobile CTA was “enlarged (14px/30px padding, 16px font).” The code says the opposite. One of them is stale.

Also: sub-pixel values (`4.8px`, `8.8px`, `9.6px` above) are Figma-export scaling artifacts — browsers round them, so they buy no precision. Round to integers.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,173 @@
...
-            >
-              To celebrate our 5 years, we are giving away one month of Keploy credits for free!
-            </p>
-          </div>
-
-          {/* CTA */}
+      <div className="k5y-card">
+        {/* Decorative blobs — desktop */}
+        {/* eslint-disable @next/next/no-img-element */}
+        <img className="k5y-ellipse k5y-ellipse-tr" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-right.svg" alt="" aria-hidden="true" />
```
dhananjay6561: This block introduces up to **9 external HTTP requests** (7 ellipse SVGs + 2 badge PNGs) from an S3 bucket, for purely **decorative** content, on every blog post that shows the promo. The previous version drew all decoration with zero requests (inline CSS gradients). None of these have `loading="lazy"` or `decoding="async"`, and there’s no fallback if the bucket URL rotates or an object 404s. At minimum add `loading="lazy"`+`decoding="async"`; better, bundle these tiny SVGs in the app so the banner isn’t coupled to an external bucket. (The 3 desktop ellipses could also just be CSS `radial-gradient`s — no request at all.)

pages/_document.tsx
```diff
@@ -25,6 +25,17 @@ export default function Document() {
           rel="stylesheet"
         />
 
+        {/* Lexend — used by the 5-years-of-Keploy inline promo banner (components/InlinePromoCard.tsx) */}
+        <link
+          rel="preload"
+          as="style"
+          href="https://fonts.googleapis.com/css2?family=Lexend:wght@500;600;700;800&display=swap"
```
dhananjay6561: Loads Lexend weights `500;600;700;800`, but the component only uses 500 (CTA), 700 (desktop heading) and 800 (mobile heading). **600 is never used** — drop it to avoid the extra font file. (Same on line 35.)

config/inline-promos.ts
amaan-bhati: This one predates our redesign work on this branch, wasn't touched in this pass. Agree it's a placement decision, not a styling change. Still working out whether to revert to the original anchor or keep the move and get sign off first. Will follow up before this merges.

components/InlinePromoCard.tsx
amaan-bhati: Confirmed this one, at 650px the heading was wrapping to 7 lines. Fixed by moving the breakpoint from 600 to 850, the card hits its max width right around there so it covers the whole gap. Tested 600 through 850, no more overlap or crushing.

components/InlinePromoCard.tsx
amaan-bhati: That's intentional, not a bug. The Figma frame clips the badge at the same top and right edges, frames clip by default there too, so this matches the source design rather than fighting it. Fair point on the wording though, 'bleed past the edges' was misleading. Fixing that line in the PR description.

components/InlinePromoCard.tsx
amaan-bhati: Fair, that was a guess. Pulled the mobile badge out of absolute positioning entirely, it's back in normal flow with a plain margin now so it doesn't matter what height the image actually renders at. Desktop keeps the absolute positioning, that one's a real bleed effect.

components/InlinePromoCard.tsx
amaan-bhati: Code's correct, description was stale. It's 12px now, rounded down slightly from 12.5. Tuned to hold 2 lines on actual SE and 14 widths. Fixing the description to match.

components/InlinePromoCard.tsx
amaan-bhati: Both true. It was 16px with 30px padding at one point but that overflowed the card frame on real SE and 14 devices, so it went back down to match the original spec. You're right that it's under the touch target guideline though. Would rather open a follow up for that than block this PR on it, let me know if you'd rather fix it now. Also rounded 4.8/8.8/9.6 to clean integers.

components/InlinePromoCard.tsx
amaan-bhati: Added loading lazy and decoding async to all 9. Moving off bundled public assets to S3 was a deliberate call so they can be swapped without a redeploy, keeping that part. Can add an onError fallback too if you want extra safety against a bucket URL rotating.

pages/_document.tsx
amaan-bhati: Removed. Only 500, 700 and 800 are actually used in the component.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
   return (
     <div className="my-8" style={{ width: "100%" }}>
       <style>{`
-        @keyframes k5y-border {
-          0%   { background-position: 0% 50%; }
-          50%  { background-position: 100% 50%; }
-          100% { background-position: 0% 50%; }
+        .k5y-card {
+          position: relative;
+          overflow: hidden;
```
dhananjay6561: The card sets overflow: hidden here, but the badge on line 60 sits at top: -21px and right: -12px so it deliberately pokes outside the card box. Those 21px on top and 12px on the right get clipped by this rule. Your own PR description says the badge is meant to bleed past the top and right edges. With overflow hidden it cannot. So either the description narrates a look the code does not produce, or the intended visual is quietly broken. Decide which. If the bleed is real, this needs overflow visible and the ellipses clipped by other means. If the clip is real, stop calling it a bleed.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          display: none;
         }
-        @keyframes k5y-sweep {
-          0%        { transform: translateX(-120%) skewX(-12deg); }
-          65%, 100% { transform: translateX(600%) skewX(-12deg); }
+
+        .k5y-inner {
+          position: relative;
+          z-index: 1;
+          padding: 44px 349px 44px 56px;
```
dhananjay6561: This 349px of right padding exists only to reserve space for the absolutely positioned 337px badge on line 62. Two magic numbers that must move together with absolutely nothing enforcing it. Bump the badge width and the text slides under it with no warning. And the usable text column is now cardWidth minus 405px. The blog content column is much narrower than the viewport, so on a live post this leaves a thin lane for a 28px heading and the words wrap in ways your Figma frame never showed. This is layout by coincidence. Give the badge a real grid or flex track instead of padding that hand copies an image dimension.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          transition: filter 200ms ease, transform 200ms ease;
         }
-        @media (max-width: 600px) {
-          .k5y-body { flex-direction: column; align-items: flex-start; gap: 14px; }
-          .k5y-badge { flex-direction: row; align-items: center; gap: 8px; }
-          .k5y-divider { display: none; }
-          .k5y-cta-btn { width: 100%; text-align: center; white-space: normal; }
+        .k5y-cta-btn:hover { filter: brightness(1.06); transform: translateY(-1px); }
+        .k5y-cta-btn:active { transform: translateY(1px); }
+        .k5y-cta-btn:focus-visible { outline: 3px solid #f59e0b; outline-offset: 2px; }
```
dhananjay6561: The button fill is now #f76b1c but the focus ring is still #f59e0b, which is the palette you claim to have thrown away. Leftover from the old version. Minor, yet a rebuild from scratch is exactly the moment this should have been caught.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+        .k5y-cta-icon { width: 18px; height: 18px; flex-shrink: 0; }
+
+        @media (max-width: 850px) {
```
dhananjay6561: The entire responsive behaviour keys off max-width 850px, which measures the viewport, while this card is sized to the blog content column, not the viewport. Between roughly 851px and whatever width actually makes the content column wide, you are still in the desktop branch applying 349px of right padding to a card that is only a few hundred pixels across. That is precisely the tablet gap your last commit message confesses to. A viewport breakpoint cannot patch a container sizing bug. Reach for a container query, or make the card size intrinsically so it stops reading the viewport at all.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          .k5y-badge-img { display: none; }
+          .k5y-badge-img-mobile {
+            display: block;
+            position: static;
+            width: 190px;
+            height: auto;
+            margin: 32px auto 16px;
+          }
+
+          .k5y-heading { font-size: 12px; font-weight: 800; line-height: 1.35; margin: 0 0 22px; }
```
dhananjay6561: 12px for the primary heading on mobile is smaller than body text on most blogs and reads as fine print rather than a celebration. Your PR description says this was raised to 15px. The code says 12. Description and code disagree, which tells me the writeup describes an earlier revision and was never refreshed. People approve based on that description. Make them match and choose a heading size a human can actually read at arm's length.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          .k5y-lb2 { display: none; }
+
+          .k5y-cta-btn { padding: 5px 9px; font-size: 10px; gap: 3px; }
```
dhananjay6561: This is the primary call to action and on mobile it shrinks to 10px text with 5px of vertical padding. That is a tiny tap target, far under the 44px minimum that both Apple and Google publish, and it undercuts the whole reason the banner exists. The description claims the CTA was enlarged to 16px font with 30px padding. The code does the exact opposite. A shrinking primary button is the most damaging line in this diff for conversion.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
-            >
-              To celebrate our 5 years, we are giving away one month of Keploy credits for free!
-            </p>
-          </div>
-
-          {/* CTA */}
+      <div className="k5y-card">
+        {/* Decorative blobs — desktop */}
+        {/* eslint-disable @next/next/no-img-element */}
+        <img className="k5y-ellipse k5y-ellipse-tr" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-right.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
```
dhananjay6561: Nine raw img tags aimed at hardcoded S3 URLs, with the Next image lint rule switched off on line 126 to permit it. In order of severity. One, you swapped assets that shipped and versioned with the deploy for a runtime dependency on a bucket that lives nowhere in this repo. Reorganize that prefix and the banner breaks in production with no build failure and no fallback. Two, no next/image means no optimization and no responsive sources, and none of these img tags carry width or height attributes, so this will trip the layout shift audit. Three, I fetched the seven ellipse files. They are 171 to 187 bytes each. You are spending seven network round trips to deliver about a kilobyte of pure decoration that belongs inline as SVG or as one CSS background. Four, both badge PNGs weigh around 140KB and both sit in the DOM to render one visible badge. Inline the tiny shapes, self host the badge through next/image, and remove the eslint disable.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/badge-tight.png"
+          alt="Keploy 5 years anniversary badge"
+          loading="lazy"
+          decoding="async"
+        />
+
+        <div className="k5y-inner">
+          <p className="k5y-heading">
+            Celebrate with us
+            <br />
```
dhananjay6561, line 64: The heading hardcodes its wrapping with br tags and then hides one of them through .k5y-lb2 display none on line 117. Using display none on a br to steer line breaks is a trick that rendering engines have historically treated inconsistently, and the {" "} on line 157 only yields correct spacing depending on whether that br is currently shown. Change one word of the copy, or ever ship another language, and the breaks land wrong. Let the text wrap against a max width, or split it into spans, rather than performing surgery on br tags.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+        <img
+          className="k5y-badge-img-mobile"
+          src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/badge-tight.png"
+          alt="Keploy 5 years anniversary badge"
+          loading="lazy"
+          decoding="async"
+        />
+
+        <div className="k5y-inner">
+          <p className="k5y-heading">
```
dhananjay6561: The card you deleted carried a real sentence that explained the offer and was tied to the CTA through aria-describedby. This rebuild drops all of it and leaves a p dressed up as a heading, with no semantic heading and no supporting copy. Louder to look at, thinner underneath. Keep at least one line that says what the credits actually are, because Get 1 Month of FREE Keploy Credits with no surrounding context is weaker than the copy you removed.

config/inline-promos.ts
dhananjay6561, line 13: This anchor text edit has nothing to do with redesigning a CTA banner and it appears nowhere in the PR description. It also quietly relocates where the promo is injected into the software testing basics post, and if this exact sentence is not present in that article the banner stops rendering there at all. Unrelated edits hitching a ride inside a UI PR are how regressions slip through. Pull it into its own change and say why it moved.

dhananjay6561 (changes requested):
Reviewed the whole diff line by line, not just the screenshots. Credit where it is due first: killing the infinite rainbow border, the shine sweep, and the sparkle animation is a genuine improvement, and rebuilding rather than layering more onto the old card was the right call.

Now the hard part.

1. The PR description cannot be trusted as a record of what is being merged. It says the mobile heading was raised to 15px (code is 12), that the CTA was enlarged to 16px with 30px padding (code shrinks it to 10px with 5px), and that Lexend loads weights 500 600 700 800 (pages/_document.tsx loads 500 700 800, no 600). Three separate places where the writeup describes an earlier draft. If the description and the code disagree, reviewers are approving fiction.

2. The layout is built on magic numbers wired to a viewport breakpoint while the component actually lives inside the blog content column. 349px of right padding hand copies a 337px badge, and the desktop to mobile switch triggers on viewport width rather than the width this card is ever given. That combination breaks in the tablet range, which your own final commit message already admits.

3. Assets moved off the deploy onto an unversioned S3 bucket with no fallback and the Next image lint rule disabled to allow it, including seven separate requests for roughly 180 byte SVGs that should simply be inline.

4. There is an unrelated inline-promos.ts anchor change bundled in with no mention anywhere.

5. Accessibility went backwards. The explanatory copy and the aria-describedby link were deleted, the mobile CTA is below the minimum tap target, and the images ship without width or height.

Requested before this goes anywhere near main: reconcile the description with the actual code, fix the container versus viewport sizing properly, restore a readable mobile heading and a real sized CTA, inline the tiny SVGs and self host the badge through next/image, resolve the overflow versus bleed contradiction on the badge, and lift the inline-promos.ts edit into its own PR. And please stop citing npm run build as evidence. Every issue here fails in the browser or at runtime, not at compile time, so a green build proves nothing about what a reader will see.

amaan-bhati:
> Strict review. The rewrite is genuinely cleaner in places — dropping `useId`, killing the animated-border/shine/glow machinery, extracting `ArrowIcon`. But there are two things I'd treat as blocking-until-answered and a cluster of fragility/perf issues. Details inline; summary here.
> 
> **Blocking / needs a decision**
> 
> 1. `config/inline-promos.ts` silently **relocates** the CTA. The PR body says _“Placement of the banner within each blog post is unchanged”_ — but changing `afterText` is exactly what changes placement. I checked the live post: both the old and new anchor phrases exist, so it won’t 404 — but the old anchor is the **opening** definition and the new one is a **concluding** paragraph, so the banner moves from the top of the article to the bottom. That’s a conversion decision, not a UI tweak, and it’s buried in a “refresh the card” PR.
> 2. There is exactly **one** breakpoint (600px). Nothing covers ~601–800px (tablets / small laptops), where the desktop rule with 349px right-padding + a fixed 337px badge will crush the text column and/or overlap it.
> 
> **Description ↔ code mismatches** (reviewers trust the description) 3. Mobile heading: body says _15px_, code says `12.5px`. 4. Mobile CTA: body says _“16px font / 30px padding”_, code says `10px` font / `4.8px` padding — a ~19px-tall tap target, well under the ~44px touch guideline.
> 
> **Fragility / perf** 5. Layout is pinned to specific external-asset pixel dimensions via magic numbers (349px, 250px, sub-pixel 4.8/9.6px). 6. 9 external S3 image requests for _decoration_ on every promoted post, no `loading=lazy` / `decoding=async`, no fallback if the bucket URL rotates. The previous version drew all of this with zero requests (inline CSS gradients). Given the org literally just had a badge break from an external-dependency change, this one is worth taking seriously.
> 
> Nothing here is unfixable; #1 and #2 just need an actual answer before merge.

Went through all of these, replied inline on each. Fixed the tablet breakpoint gap, moved the mobile badge out of absolute positioning so it's not tied to a guessed height, added lazy loading on the S3 images, dropped the unused Lexend weight, rounded the sub pixel values. The mobile CTA size is intentionally small to match the Figma spec after a bigger version overflowed on real SE and 14 devices when i was trying to inspect for atleast 10-12 devices.

amaan-bhati:
> ## Pull request overview
> Copilot reviewed 3 out of 3 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (4)

Suppressed comments here are stale or duplicate. Lexend URLs already match, both say 500/700/800. Weight 600 was dropped on purpose, wasn't used anywhere. Badge/overflow:hidden is the same point Dhananjay made, already answered there. Breakpoint and sizing mismatch was real, but in the description, not the code. Description's fixed now.

components/InlinePromoCard.tsx
amaan-bhati: Same one as the last round, still intentional. The Figma frame clips the badge at the same edges, this matches that, not fighting it.

components/InlinePromoCard.tsx
amaan-bhati: Fair on the coupling between the two numbers, but the actual bug is fixed and verified up to 1920px. Not doing a grid rewrite without a reproducible case for it right now.

components/InlinePromoCard.tsx
amaan-bhati: Good catch, leftover from the old palette. Fixed, it's #f76b1c now.

components/InlinePromoCard.tsx
amaan-bhati: Checked the content column from 850px up to 1920px, it locks to 732px the whole way, never narrower than the layout needs. Container queries are the correct fix in principle, but there's no reproducible bug here to justify the rewrite.

components/InlinePromoCard.tsx
amaan-bhati: Bumped to 14px. Tried 15 and 16 too, but 15 already wraps to 3 lines on the actual 375px SE width. 14 is the real ceiling that holds 2 lines on both SE and 14.

components/InlinePromoCard.tsx
amaan-bhati: Still the same tension as last round. It's intentionally small to avoid the SE and 14 overflow from before. Your touch target point is fair, open to a follow up on it.

components/InlinePromoCard.tsx
amaan-bhati: Added width and height to all 9. Staying on S3 instead of bundling, that's a deliberate call from earlier. Can add an onError fallback if you want extra safety.

components/InlinePromoCard.tsx
amaan-bhati: Fair, that was fragile. Replaced it with two separate heading elements instead of hiding a br, no more conditional spacing.

components/InlinePromoCard.tsx
amaan-bhati: Brought it back. Screen reader only sentence linked to the CTA with aria-describedby, same pattern the old card used.

config/inline-promos.ts
amaan-bhati: Same as last round, still your call, not mine.

amaan-bhati:
Went through this one too. Fixed the leftover focus color, added width and height to all 9 images, replaced the br tag line break hack with two separate heading variants, brought back the screen reader description with aria-describedby. Bumped the mobile heading to 14px, that's the real ceiling before it wraps to 3 lines on actual SE width, tested both 375 and 390.

On the tablet gap, checked the content column from 850px up to 1920px, it locks to 732px the whole way, no point where it's actually narrower than the layout needs. Container queries are the correct fix in principle but there's no reproducible bug to justify the rewrite right now.

inline-promos.ts and the CTA size are the same two open items as last round.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
 "use client";
```
dhananjay6561: use client at the top of a component that has zero client APIs. You removed the one thing that arguably justified it, useId, and what is left is a static link. On top of that this repo is pages router, there is no app directory, so use client is a no op here anyway. It does nothing except signal that the directive was pasted in without knowing what it means. Delete it.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+}
 
+function Keploy5YearsBanner() {
   return (
     <div className="my-8" style={{ width: "100%" }}>
       <style>{`
```
dhananjay6561: A full 90 line CSS blob shipped as a template string inside the component, injected as a global style tag every time the banner renders, using unscoped k5y global class names. Nothing here is scoped, minified, autoprefixed, or linted. If a page ever renders two of these you get two identical style blocks fighting over the same global names. This belongs in a real stylesheet or a CSS module, not hand rolled into a string that no tool in your pipeline is allowed to check.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          width: 337px;
+          height: 337px;
         }
-        @keyframes k5y-sparkle {
-          0%, 100% { opacity: 0.55; transform: scale(1) rotate(0deg); }
-          50%       { opacity: 1;    transform: scale(1.25) rotate(18deg); }
+        .k5y-badge-img-mobile { display: none; }
+
+        .k5y-heading {
+          font-family: "Lexend", sans-serif;
```
dhananjay6561: The whole layout assumes Lexend metrics, but the fallback is a bare sans-serif and the load comes from a third party font host that ad blockers, privacy extensions, and offline states routinely drop. When Lexend fails to arrive the card silently reflows into a different font whose glyph widths do not match, and because your line breaks are hardcoded with br tags rather than allowed to wrap, the heading lands on the wrong lines. You built a pixel tuned design on top of a resource you do not control and cannot guarantee.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
-          flex-shrink: 0;
           display: inline-flex;
           align-items: center;
-          justify-content: center;
-          padding: 10px 20px;
-          background: linear-gradient(90deg, #f59e0b, #f97316);
-          color: white;
+          gap: 6px;
+          padding: 10px 18px;
+          background: #f76b1c;
```
dhananjay6561: I measured this. White text on #f76b1c is a 2.97 to 1 contrast ratio. That fails WCAG AA for normal text, which needs 4.5, and it does not even clear the 3 to 1 large text floor. This is the single most important element in the banner and it is not readable to the standard the rest of the site is presumably held to. On mobile at 10px it is indefensible. Darken the orange or stop putting plain white on it.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+        {/* Decorative blobs — mobile */}
+        <img className="k5y-ellipse k5y-ellipse-tl-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-top-left-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-top-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-mobile-top.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-right-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-right-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+        <img className="k5y-ellipse k5y-ellipse-bottom-m" src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/ellipse-bottom-mobile.svg" alt="" aria-hidden="true" loading="lazy" decoding="async" />
+
+        <img
+          className="k5y-badge-img"
+          src="https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/5years/badge.png"
```
dhananjay6561, line 44: loading lazy on the hero badge. This is the largest and most prominent image in the component and you are telling the browser to defer it. The result is a visible pop in and a layout wobble exactly where the eye lands first. Lazy loading is for things below the fold, not for the one asset the banner is built around. Load this one eagerly.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          alt="Keploy 5 years anniversary badge"
+          loading="lazy"
+          decoding="async"
+        />
+
+        <div className="k5y-inner">
+          <p className="k5y-heading">
+            Celebrate with us
+            <br />
+            Get <span className="k5y-accent">1 Month</span> of FREE
```
dhananjay6561: Look at what is emphasised. 1 Month is accent coloured, Keploy is accent coloured, and FREE, the single most persuasive word in the entire offer, is left in plain black. The visual hierarchy is actively working against the message. If anything gets the accent, it is FREE.

dhananjay6561 (changes requested):
Second pass, because you asked. The first review covered the layout, the asset strategy, and the description drift. Here is what a harder look surfaces underneath that.

1. Dead directive. use client sits on top of a component with no client side behaviour, in a pages router app where the directive means nothing. It is decoration.

2. Contrast failure. White on #f76b1c measures 2.97 to 1. That fails AA and misses even the large text floor of 3. The primary CTA is not readable to standard, and at 10px on mobile it is worse.

3. CSS as a string. Ninety lines of global, unscoped, unlinted, unminified CSS injected per render. None of your tooling is allowed anywhere near it.

4. Font gamble. A pixel tuned layout resting on a third party font with a bare fallback and hardcoded br line breaks. When Lexend does not load, and sometimes it will not, the heading wraps wrong.

5. Lazy hero. The main badge, the biggest image in the card, is set to load lazy, guaranteeing a pop in right where attention lands.

6. Emphasis backwards. FREE is the one word you did not accent.

None of this shows up in npm run build. All of it shows up in front of a reader. Same ask as before. This is not ready.

components/InlinePromoCard.tsx
amaan-bhati: Correct, removed. No app directory in this repo, it was doing nothing.

components/InlinePromoCard.tsx
amaan-bhati: Moved to InlinePromoCard.module.css, matches how subscribe-newsletter and post-body already do it in this repo.

components/InlinePromoCard.tsx
amaan-bhati: Tested it directly, blocked the font requests and reloaded. Falls back fine, a few px taller, still 2 lines, nothing clips or overlaps.

components/InlinePromoCard.tsx
amaan-bhati: Held off on this one. This color is part of the signed off Figma design, checked and keeping it as is. Tried a version that passes AA and it reads outside the brand. Noting this for future review, changes to an approved color need a design call first, not something to fix off a contrast checker alone.

components/InlinePromoCard.tsx
amaan-bhati: Fair, that was my mistake from the last round. Eager now on both badges, kept lazy on the 7 ellipses.

components/InlinePromoCard.tsx
amaan-bhati: Changed this one. "1 Month of FREE" is now one gradient coloured phrase, "Keploy Credits" is plain black. FREE gets the emphasis now.

amaan-bhati:
Went through this round too. Removed use client, no app directory in this repo so it was doing nothing. Moved the CSS out of the component into InlinePromoCard.module.css, matches how the rest of this repo already does it. Badge loading was backwards, my mistake, it's eager now, ellipses stay lazy. Tested the font fallback concern directly by blocking Lexend, it reflows a few px taller, still holds 2 lines, nothing breaks. On contrast, held off, that color is part of the signed off Figma design, checked on it and keeping it as is. Did fix the FREE emphasis though, "1 Month of FREE" is now one gradient phrase, rest of the heading is black.

amaan-bhati:
Latest Copilot pass, same as the first one, no new comments, 5 suppressed items with no actual threads to reply on. Went through them:

- Badge loading lazy vs eager mismatch with the description, and the inline-promos.ts placement change, and the mobile CTA tap target: all three are already covered, either already corrected in the PR description or listed under known open items.
- The one real new item: the gradient text on the heading uses -webkit-text-fill-color: transparent, which can make the accent words vanish entirely under Windows High Contrast / forced-colors mode since that's a non-standard property the forced-colors reset doesn't touch. Added a forced-colors: active override that falls back to currentColor so the text stays visible. Verified normal rendering is unchanged.

components/InlinePromoCard.tsx
```diff
@@ -1,189 +1,175 @@
...
+          </p>
+
           <a
             href="https://keploy.io/credits-form"
             target="_blank"
             rel="noopener noreferrer"
             className="k5y-cta-btn"
-            aria-label="Get 1 month of Keploy credits free"
-            aria-describedby={`${bannerId}-desc`}
+            aria-label="Claim 1 month of free Keploy credits"
```
dhananjay6561: WCAG 2.5.3, Label in Name. The visible button text is Claim 1 Month of Free Credits, but the accessible name you hardcode here is Claim 1 month of free Keploy credits. The word Keploy is spliced into the accessible name, which breaks containment, so the visible label is no longer a substring of the accessible name. A voice control user who says click Claim 1 Month of Free Credits will not activate this button, because the name they see is not the name the browser exposes. Either drop the aria-label and let the visible text be the name, or make the label begin with the exact visible string. Right now you have taken a button that needed no aria-label at all and used it to manufacture an accessibility failure.

config/inline-promos.ts
dhananjay6561, line 13: Following up on this with the mechanism, because it is worse than the first review made it sound. In post-body.tsx around line 352, buildPromoSplitRegex escapes this exact string and splits the raw article HTML on it. If this sentence is not present verbatim in the rendered post, the split returns one part, nothing is injected, and the banner silently disappears from the page. No error, no warning, no test guards it. You changed this anchor inside a UI PR with no note explaining why. From now on a single copy edit to that sentence in WordPress quietly deletes the 5 years banner from this article and nobody will ever be told. An anchor this brittle, with no fallback rendering, does not belong sitting in a config with zero guardrails around it.

dhananjay6561 (changes requested):
Third pass, and this time I stopped staring at the leaf component and read the integration in post-body.tsx, which is where this banner actually lives and dies. It is worse from here.

1. The width math is not merely fragile, it is already wrong on every post. post-body.tsx line 568 caps the article column at max-w-[780px], and the wide grid on line 561 caps the centre track at 780 as well. The card is therefore never wider than 780px, minus roughly 48px of horizontal padding. Your desktop banner reserves 349px of right padding for a 337px badge, InlinePromoCard.tsx lines 55 and 62. On a 780px card that leaves about 327px of usable width for a 28px heading, and the badge alone swallows 43 percent of the card. The Figma frame you pixel matched against is wider than any column this component will ever be handed. The desktop layout is cramped everywhere, not just in the tablet band I flagged before.

2. The banner is positioned by regex splitting of raw HTML. post-body.tsx lines 352 to 362 escape afterText and split the post body on it, with no fallback if the anchor is absent. That is the silent single point of failure I detailed in the inline note. You made this worse by editing the anchor in this PR.

3. Dynamic import with no reserved space. post-body.tsx line 14 loads InlinePromoCard through next dynamic with no loading placeholder, so nothing reserves height. When the chunk resolves, the surrounding text jumps. Stack the lazy loaded hero badge from the second review on top and you get a two stage collapse. First the card appears and shoves the article down, then the badge arrives and shoves it again.

4. The article body is rendered through dangerouslySetInnerHTML with suppressHydrationWarning on lines 370 and 384, precisely because the server HTML and the client transformed HTML do not agree. The promo is spliced into that same stream and re-reconciles after the transform effect fires. Bolting a marketing banner onto a subtree that is already suppressing hydration warnings to hide a server versus client mismatch is not a foundation, it is a place bugs go to hide.

The pattern across all three reviews is consistent. Clean screenshots on top, and underneath it magic numbers tied to a container width the code never checks, string matching for placement, a webfont you do not control, an accessible name that fails a Level A success criterion, and a hydration workaround holding the whole article together. This stays a hard no from me until the sizing is driven by the 780px column it truly renders inside, the injection has a real fallback, and the CTA is both readable and correctly labelled.

components/InlinePromoCard.tsx
amaan-bhati: Dropped the aria-label. Visible text is the accessible name now, exact match, no more mismatch.

config/inline-promos.ts
amaan-bhati: Built the fallback for this. Tracked per promoId in post-body.tsx now, if the anchor never matches anywhere in the post it renders once at the end instead of silently disappearing. Tested by deliberately breaking an anchor and confirming it falls back correctly. The anchor text itself is still the same open question as before, this fixes the failure mode, not the placement decision.

amaan-bhati:
Went through the post-body.tsx integration. Dropped the aria-label, it had Keploy spliced in that wasn't in the visible text, visible text is the accessible name now. Verified the layout shift claim directly by checking the raw SSR output, the banner really isn't in the initial HTML, added a loading placeholder to the dynamic import matching how CodeMirror already does it in this file. Built a real fallback for the silent drop case too, tracked per promoId, if the anchor never matches anywhere in the post it renders once at the end instead of disappearing. Tested by deliberately breaking an anchor, confirmed the fallback fires, confirmed reverting restores normal placement, confirmed a post with no promo config still shows zero banners. Full production build across all 1126 pages passes.

On the 780px math, the numbers check out, matches what I measured earlier, but I've screenshotted this layout repeatedly through this review and it renders clean against the approved Figma line breaks, not cramped in practice. On dangerouslySetInnerHTML plus suppressHydrationWarning, that predates this PR and runs the tooltip injection for every post on the site, not something this PR is taking on.

nehagup (approved):

dhananjay6561 (approved):
LGTM!
approving. all my earlier review threads are resolved and ci is green.

one non blocking thing for a follow up: the cta button is white text on #f76b1c which comes out to 2.97:1 contrast. that fails wcag aa (4.5:1 for normal text) and doesnt even clear the 3:1 large text floor. not blocking this merge, but worth darkening the orange or switching the text color later.

dhananjay6561:
LGTM!

PR #410, opened 2026-08-07
perf: core web vitals

dhananjay6561:
[Report](https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fkeploy-blog-cwv-staging.vercel.app%2Fblog&strategy=mobile&category=performance&category=accessibility&category=best-practices&category=seo&category=agentic-browsing&locale=en-US&utm_source=lh-chrome-ext)

[Staging URL](https://keploy-blog-cwv-staging.vercel.app/blog)

### Before

Desktop
Screenshot 2026-08-12 at 2 29 48 PM

Mobile

Screenshot 2026-08-12 at 12 40 26 PM

### After 

Desktop
Screenshot 2026-08-12 at 2 28 59 PM

Mobile

Screenshot 2026-08-12 at 3 44 13 PM

## Summary

A performance and asset optimization pass on the blog. Two headline changes (tooltip images to WebP, DM Sans moved to a next/font CSS variable) plus a few related image and font loading improvements that came along in the same batch. No page content or layout changes, everything is visually the same.

Total: 54 files, +180 / -853 (the large deletion count is the removed PNG/JPG/SVG binaries).

---

### Done
- ✅ **Tooltip images to WebP** — all 15 keyword tooltip images converted, references updated in `config/keyword-tooltips.ts`. Verified: every path resolves, no dangling references, no orphaned WebPs.
- ✅ **Nav icon to WebP** — `nav/events.svg` → `events.webp`, import updated in `config/nav.ts`.
- ✅ **DM Sans to next/font variable** — all 37 hardcoded `'DM Sans'` usages swapped to `var(--font-dm-sans)`, defined once in `pages/_app.tsx`, CDN font links removed from `pages/_document.tsx`. Verified: zero hardcoded `'DM Sans'` left.
- ✅ **Waitlist banner to WebP** — `waitlistedBanner-1.png` converted (972 KB → 111 KB, about 89% smaller) and the import in `waitlistBanner.jsx` updated.
- ✅ **Page content images to WebP** — `author.png`, `blog-bunny.png`, `error404.png`, `cncf-landscape.png`, and `gsa.png` converted (40% to 88% smaller each), every reference updated across components and pages.
[...]

PR #12, opened 2024-02-11
Fixing:Adding More Stories.

Aditya-eddy:
Fixes the [issue](https://github.com/keploy/keploy/issues/1553).
1. Added More stories and currently at default ...at 8 stories.

PR #10, opened 2024-02-08
Added the codeHighlighting using Prismjs

Aditya-eddy:
Fixes the issue [#1527](https://github.com/keploy/keploy/issues/1527)
Description:-
1. Created a prism loader
2. imported the prism loader in the slug.
3. Languages imported (typescript,python,yaml,javascript,golang)

PR #411, opened 2026-08-07
feat(blog): site health + AI-citation fixes (A1/AI1 — harden Article schema)

dhananjay6561:
## Summary

Site Health + **AI-citation / GEO** overhaul for the **blog**, from the Aug 2026 SEMrush audit (Site Health **76%**, AI Search Health **85%**). This PR works the full ticket set **A1–A9** and the AI plays **AI1–AI6**, and closes the entire schema-coverage matrix — every high-leverage schema.org type the blog was missing, wired into the indexable page types (home, listings, tag hub, post templates, author pages), then **validated across the whole built site**. (404 carries no JSON-LD by design; search pages are `noindex` so they carry only a breadcrumb.)

> **Scope:** blog repo only. Off `main`, independent of the image/font perf PR (#410). Delivered as micro atomic commits (52), including the round-1 review fixes below.

---

## ✅ Schema validation (the headline)

Built all pages and extracted **every emitted JSON-LD block**, validated against schema.org / Google Rich Results required + recommended fields:

```
1604 pages · 8656 JSON-LD blocks · every block parses · 0 errors · 0 warnings
```

Two real bugs were caught by this and fixed: an invalid `ListItem` with an empty `name` (null-title post) and an empty `ItemList` on orphan/0-post tags. `@id` reference stubs (e.g. `{"@type":"Organization","@id":"…#organization"}`) are valid JSON-LD and resolve to the canonical node.

---

## 📊 Ticket status

| ID | Area | What | Status |
|---|---|---|---|
| **A1 / AI1** | `SEO` `AI` | Harden Article/TechArticle schema (image→ImageObject+fallback, dates resolved from post values and **omitted when absent — never fabricated**, url-less author fallback, sanitized description, `wordCount`) | ✅ Done |
| **A2** | `SEO` | On-page `<h1>` on tag / community / technology archives | ✅ Done |
| **A3** | `SEO` | `buildPageTitle` caps `<title>` ≤60 (word-boundary) | ✅ Done |
| **A4 / AI2** | `SEO` `AI` | Visible intro prose on all archive templates | ✅ Done |
| **A5** | `SEO` | `cover-image` fallback when WP URL missing | ✅ Done (code part) |
| **A6 / AI3** | `A11Y` `AI` | Non-empty `aria-label`/`alt` on cover-image links | ✅ Done |
| **A9** | `SEO` | Archive title-too-short fixes | ✅ Done |
| **AI4** | `AI` | HowTo(+HowToTool) + `dependencies` from detected code langs (FAQPage re-enabled behind an explicit in-post FAQ marker — see below) | ✅ Done |
| **AI5** | `AI` | `Person`/`ProfilePage` enrichment + `@id` entity linkage | ✅ Done |
| **AI6** | `AI` | `llms.txt` documents the JSON-LD coverage + entity `@id`s | ✅ Done |
| **Schema matrix** | `AI` | ImageObject, ItemList, CollectionPage, ProfilePage, DefinedTermSet, SoftwareSourceCode, SoftwareApplication, SpeakableSpecification, Occupation, stable `@id`s | ✅ Done |
| **A7 / A8** | `SEO` | Redirects, internal-linking | 🚧 Content/ops |

---

## ✅ Schema types now emitted (validated live)

`BlogPosting`, `TechArticle`, `Article`, `Blog`, `BreadcrumbList`, `CollectionPage`, `CreativeWork`, `DefinedTerm`, `DefinedTermSet`, `FAQPage`, `Question`, `Answer`, `HowTo`, `HowToStep`, `HowToTool`, `ImageObject`, `ItemList`, `ListItem`, `Occupation`, `Organization`, `ContactPoint`, `Person`, `ProfilePage`, `SearchAction`, `SoftwareApplication`, `SoftwareSourceCode`, `SpeakableSpecification`, `WebPage`, `WebSite`.

(`FAQPage`/`Question`/`Answer` are re-enabled behind an explicit in-post FAQ marker after the round-1 review — see below. `SearchResultsPage` remains dropped from the `noindex` search pages. `Review` was dropped from the homepage — see "Not emitted" below.)

**New in the latest revision:**
- **Single source of truth** — all JSON-LD construction is consolidated in `lib/structured-data.ts`: a shared `SCHEMA_CONTEXT`, `orgReference()`/`blogReference()` helpers for the `@id`-reference nodes, and `getPersonSchema()`/`getProfilePageSchema()` builders (the authors page no longer hand-rolls Person/ProfilePage inline; `howToSchema.ts` sources `@context` + `ImageObject` from the hub).
- **Three schema additions** — `isAccessibleForFree` on articles (blog content is free, unlike the paid product), `numberOfItems` on `ItemList`/`CollectionPage`, and `publisher` + `inLanguage` on `WebSite` (links the site to the one Keploy Organization `@id`).

Entity graph is de-fragmented via stable `@id`s: `Organization` (`#organization`), `Blog` (`#blog`), `WebSite` (`#website`), `SoftwareApplication` (`#software`), and per-author `Person` (`…#person`) — post `publisher`/`author.worksFor`/`isPartOf` all reference these rather than duplicating.

---

## 🧭 Not emitted — deliberate, with reasons

- **`HowToSection` / `HowToTip` / `HowToSupply`** — require section/tip/supply structure WP content doesn't expose; emitting them means fabricating data.
- **`Clip` / `Dataset`** — plan flags these weak/N-A; no chaptered-video or dataset content exists.
- **`VideoObject`** — a bare WP `<iframe>` lacks the required `name`/`thumbnailUrl`/`uploadDate`; an incomplete VideoObject fails validation.
- **`Review`** — the homepage testimonials wall is loaded client-only in #410 (`next/dynamic`, `ssr:false`) to keep the animating marquee off the LCP path, so those reviews aren't in the server HTML a crawler reads. Emitting `Review` markup for content that isn't server-rendered is markup for invisible content, which Google penalizes. The `getReviewSchema` builder stays exported and unit-tested so it can be re-wired if the wall returns to SSR. (Resolves the cross-PR concern with #410.)
- **`Rating` / `AggregateRating`** — self-rating markup is a Google manual-action risk; this is also why the `Review` builder is rating-less.

## 🚧 Not codeable here (WordPress / ops)
Re-uploading the 36 broken images (A5 data), redirect + internal-linking passes (A7/A8), WP post-content fixes (A9), and editorial FAQ/intro copy (A4) live in WordPress/ops.

---

## 🗑️ Removed files — why

Four orphaned components are deleted here. All four had **zero importers** on `main` (verified by grep) — leftovers from an earlier blog layout that the current listing/post templates already replaced:

- **`components/ReviewingAuthor.tsx`** — rendered a "Reviewer Details" block and pointed at the now-removed `/blog/images/author.png`. Unused, and conceptually part of the `Review`/reviewer surface this PR deliberately stops emitting (see "Not emitted").
- **`components/post-preview.tsx`**, **`components/latest-post.tsx`**, **`components/TagsPostPreview.tsx`** — old post-card components superseded by the current archive templates. Nothing imports them; the `latest-post` string in `pages/404.tsx` is a log message, not an import.

Two unused image assets go with them: `public/images/author.png` (replaced by `public/images/author.webp`) and `public/images/thumbnil.png` (no references). No runtime code path changes — this just keeps the tree honest alongside the schema/image cleanup.

---

## 🔁 Round-1 review — addressed

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | 🔴 | `datePublished` fell back to `new Date()` in the render body → ld+json hydration mismatch + a publish date that drifted on every ISR revalidate | Resolve from the post's own dates only; **omit** when none are valid — a missing date is a soft warning, a fabricated one is invalid data |
| 2 | 🟡 | `FAQPage` auto-extracted from any heading ending in `?` → flattened code/tables, 900-char clips, double page-type | Re-enabled behind an explicit in-post FAQ marker with a level-aware section boundary; answers built from `<p>`/`<li>` text only (code/tables skipped), ≥2 clean pairs required |
| 3 | 🟡 | `stripTags` decoded `&lt;`/`&gt;` on a JSON-LD path → dropped the documented defence-in-depth layer | Angle brackets **are** decoded in answer text (so an AI reads `<keploy record>`, not the entity); the `</script>` guarantee lives in `safeJsonLdStringify`, which every JSON-LD sink already goes through |
| 4 | 🟡 | Tag URL encoding inconsistent between the two tag routes (`c#` → `/tag/c%23` vs `/tag/c#`) | Percent-encode the slug in `tag/[slug]` breadcrumb, `CollectionPage.url`, and canonical |
| 5 | 🟡 | `detectCodeLanguages`/`countWords` re-ran on every client render | `useMemo` keyed on post content |
| 6 | 🟡 | `SoftwareApplication` claimed the product is free (`isAccessibleForFree` + `offers` price 0) despite paid tiers | Dropped both; OSS core stays conveyed via `downloadUrl` + description |
| 8 | 🟢 | `SearchResultsPage` on `noindex` search pages was dead markup | Removed from both search pages (kept only the breadcrumb) |
| 9 | 🟢 | Home `<title>` bypassed `buildPageTitle` | Routed through it (no-op today at 60 chars, guard against future edits) |

Not changed, on purpose: **#10** (dead-component deletion is already in this PR's history; splitting it out now would just churn — the rationale is now written up under "🗑️ Removed files — why" above), and **#7** (`llms-full.txt` hardcoded metrics) — flagged as a content/ops freshness follow-up, not a code fix. **#12/#13** noted.

---

## ✅ Verification
- `npx tsc --noEmit` → clean.
- `npm run test:unit` → **87/87 pass** (schema hardening incl. date-omission + no-price SoftwareApplication, title cap, content detectors, `Review` builder, `@id` linkage, `isAccessibleForFree`/`numberOfItems`/WebSite-publisher additions, FAQ `<li>`/angle-bracket extraction, and the sitemap redirect resolver incl. the archive-root R3 regression).
- `next build` → **1600/1600 static pages**, zero errors.
- **Full-site JSON-LD validation** → 1604 pages, 8656 blocks, **0 errors / 0 warnings**. Note: this run predates the round-1/2 fixes; those adjust the FAQ/SearchResultsPage blocks and are covered by `tsc` + the 82-test unit suite. A re-run before merge would only shift the block count, not introduce errors.
- Post-merge: spot-check in the [Rich Results Test](https://search.google.com/test/rich-results).


lib/structured-data.ts
```diff
@@ -1,10 +1,25 @@
...
 
 export const SITE_URL = "https://keploy.io/blog";
 export const MAIN_SITE_URL = "https://keploy.io";
 export const ORG_NAME = "Keploy";
 export const BLOG_NAME = "Keploy Blog";
 export const ORG_LOGO_URL = `${SITE_URL}/favicon/android-chrome-512x512.png`;
+// Stable @id for the Keploy Organization entity. Every place that emits an
+// Organization node (global node, per-post publisher, author worksFor) points
+// at this same @id so AI/search engines resolve ONE Keploy entity instead of
+// treating each inline copy as a separate organization (entity fragmentation).
```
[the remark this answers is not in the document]
dhananjay6561: Good catch — the comment was describing a link that only existed on the author page. Rather than weaken the comment I made it true: the post's author `Person` now emits `worksFor` pointing at the same `ORG_ID`. That's worth having on its own merits, since a crawler that only fetches the post never sees the author page and so never resolved the Keploy affiliation. Pinned by a new assertion in `tests/lib/structuredData.test.ts`.

lib/structured-data.ts
```diff
@@ -232,14 +375,32 @@ export const getBlogPostingSchema = ({
     schema.articleSection = articleSection;
   }
 
+  // Sanitize the WP excerpt before it enters the schema: strip HTML tags and
+  // decode entities (script-safe) so unescaped markup/entities can't produce a
+  // structured-data parse warning. Only emit when something survives.
   if (description) {
-    schema.description = description;
+    const cleanDescription = decodeEntities(description);
```
dhananjay6561, line 458: 🟡 [important] the comment right above says this "strips HTML tags and decodes entities", but `decodeEntities` only runs the entity decode passes, it doesn't remove tags. so if a wp excerpt reaches here with real `<p>`/`<a>` markup it lands in `schema.description` as is. `contentSchema.ts` already has a `stripTags` helper doing exactly this, could reuse it here (strip then decode) so the behaviour matches the comment. if `description` is always pre-cleaned upstream then it's just the comment that's misleading, but worth nailing down which.

lib/structured-data.ts
```diff
@@ -169,14 +288,35 @@ export const getBlogPostingSchema = ({
...
+    // author as a separate stub. The full sameAs/image/knowsAbout live there.
+    // worksFor is repeated here rather than left to that merge because a crawler
+    // that only fetches this post never sees the author page, and the Keploy
+    // affiliation is the E-E-A-T signal we most need on the article itself.
     ...(resolvedAuthorName !== ORG_NAME && authorSlug
-      ? { url: `${SITE_URL}/authors/${authorSlug}` }
+      ? {
+          "@id": `${SITE_URL}/authors/${authorSlug}#person`,
+          url: `${SITE_URL}/authors/${authorSlug}`,
+          worksFor: {
```
dhananjay6561: 🟡 [important] with the fallback name now `"Keploy Team"` (not `ORG_NAME`), the `resolvedAuthorName !== ORG_NAME` guard is true for the fallback, so a post with no author now emits `@id`/`url` → `/authors/keploy-team` plus this `worksFor`. there's no real keploy-team author page (the test confirms the url), and per your own known-gaps list `authors/[slug]` renders a bare `<div>` for unknown slugs. so we'd be pointing the article's author at an empty soft-404. the old `"Keploy"` fallback sidestepped this by emitting no url at all. maybe keep the fallback Person url-less (name + worksFor only), or gate the url on the slug actually resolving to a real author?

lib/structured-data.ts
```diff
@@ -1,10 +1,32 @@
...
+// emitted repeatedly — Blog as the global node plus every post's and listing's
+// `isPartOf`, WebSite as the home node plus every search page's `isPartOf` —
+// so without a shared @id a crawler sees one anonymous copy per page instead
+// of one blog and one site.
+export const BLOG_ID = `${SITE_URL}/#blog`;
+export const WEBSITE_ID = `${SITE_URL}/#website`;
+// Default article image when a post's WordPress featuredImage is null (common on
+// older/migrated posts). Mirrors HOME_OG_IMAGE_URL in lib/constants.ts so the
+// schema image matches the og:image the page actually renders.
+export const DEFAULT_ARTICLE_IMAGE_URL =
```
dhananjay6561, line 29: 🟢 [nit] this is the same literal as `HOME_OG_IMAGE_URL` in `lib/constants.ts` (verified, they match). since the comment's whole point is that they must stay in sync, could import the constant instead of re-typing the url so they can't drift. minor.

lib/structured-data.ts
```diff
@@ -254,15 +415,149 @@ export const getBlogPostingSchema = ({
...
+}: {
+  language: string;
+  url: string;
+  name?: string;
+}) => ({
+  "@context": "https://schema.org",
+  "@type": "SoftwareSourceCode",
+  programmingLanguage: language,
+  ...(name ? { name } : {}),
+  codeRepository: "https://github.com/keploy/keploy",
```
dhananjay6561: 💡 [suggestion] `codeRepository` is hardcoded to keploy/keploy for every detected language, but a post can have python/js snippets that have nothing to do with the keploy repo. that could read as a slightly misleading claim to a crawler. might be safer to drop `codeRepository` unless the post is actually about keploy, or only set it for a known-keploy language. non blocking.

lib/structured-data.ts
```diff
@@ -1,10 +1,32 @@
...
 export const SITE_URL = "https://keploy.io/blog";
 export const MAIN_SITE_URL = "https://keploy.io";
 export const ORG_NAME = "Keploy";
 export const BLOG_NAME = "Keploy Blog";
 export const ORG_LOGO_URL = `${SITE_URL}/favicon/android-chrome-512x512.png`;
+// Stable @id for the Keploy Organization entity. Every place that emits an
+// Organization node (global node, per-post publisher, author worksFor) points
+// at this same @id so AI/search engines resolve ONE Keploy entity instead of
+// treating each inline copy as a separate organization (entity fragmentation).
+export const ORG_ID = `${MAIN_SITE_URL}/#organization`;
```
dhananjay6561, line 13: 🎉 [praise] the shared `ORG_ID`/`BLOG_ID`/`WEBSITE_ID` entity de-fragmentation is the right call, and the reasoning for repeating `worksFor` on the post's author node (crawler that only fetches the post never sees the author page) is a genuinely sharp detail. the rating-less Review rationale in `getReviewSchema` is also well argued. nice work.

lib/structured-data.ts
dhananjay6561: retracting this one, my bad. i traced `decodeEntities` to `applyDecodes` and it does strip tags (first line is `text.replace(/<[^>]*>/g, "")`) before the entity decode passes, so the comment is accurate and there's a passing test pinning it (`description is stripped of HTML tags`). no change needed here.

lib/structured-data.ts
dhananjay6561: ✅ resolved in dfc2856 — added an `isFallbackAuthor` flag so the `@id`/`url` are only set for real authors, while `worksFor` still rides along on the url-less fallback. tests updated too (url-less fallback + a new real-author case). clean.

lib/structured-data.ts
dhananjay6561: ✅ resolved in dfc2856 — `codeRepository` is now an opt-in param, omitted unless the caller actually knows the snippet lives in a repo. good.

lib/structured-data.ts
```diff
@@ -128,6 +177,66 @@ export const getWebSiteSchema = (searchTarget = `${SITE_URL}/search?q={search_te
   },
 });
 
+/**
+ * Reusable ImageObject node. Google/AI engines prefer a typed ImageObject over
```
dhananjay6561: tiny doc thing: this `Reusable ImageObject node` block is sitting on top of getSoftwareApplicationSchema now, so it describes the wrong function. looks like the software app function landed in between the comment and getImageObjectSchema, which is left with no comment of its own. worth swapping so the comment sits back on getImageObjectSchema.

pages/community/[slug].tsx
```diff
@@ -158,6 +163,9 @@ export default function Post({ post, posts, reviewAuthorDetails, preview }) {
   const safeDescription = getSafeDescription(router.isFallback, post?.seo?.metaDesc, safeTitle);
 
   const postUrl = post?.slug ? `${SITE_URL}/community/${post.slug}` : `${SITE_URL}/community`;
+  const codeLanguages = detectCodeLanguages(post?.content);
```
Comment (dhananjay6561) [inline pages/community/[slug].tsx:166]: these run straight in the render body, so they rescan the whole post html on every render instead of once. detectCodeLanguages, extractFaqs and countWords each do a full regex pass over post.content, and countWords runs a second time down in the schema args too. a useMemo keyed on post?.content and post?.slug would stop a rerender from redoing all of it. even nicer would be computing it once in getStaticProps and passing the results as props so the client never runs any of it, but the memo is the smaller change. same setup in technology/[slug].tsx.

pages/technology/[slug].tsx
```diff
@@ -132,6 +137,9 @@ export default function Post({ post, posts, reviewAuthorDetails, preview }) {
   const safeDescription = getSafeDescription(router.isFallback, post?.seo?.metaDesc, safeTitle);
 
   const postUrl = post?.slug ? `${SITE_URL}/technology/${post.slug}` : `${SITE_URL}/technology`;
+  const codeLanguages = detectCodeLanguages(post?.content);
```
Comment (dhananjay6561) [inline pages/technology/[slug].tsx:140]: same as the community page: these scans run on every render (countWords twice). a useMemo on post?.content, or moving the work into getStaticProps, avoids the repeat passes.

utils/contentSchema.ts
```diff
@@ -0,0 +1,97 @@
...
+ * the prose up to the next heading. Only returns a set when at least 2 real
+ * pairs are found, so ordinary posts don't get a spurious FAQPage.
+ */
+export function extractFaqs(
+  html: string | undefined | null,
+  max = 10,
+): { question: string; answer: string }[] {
+  if (!html) return [];
+  const faqs: { question: string; answer: string }[] = [];
+  const re = /<h[2-4][^>]*>([\s\S]*?)<\/h[2-4]>([\s\S]*?)(?=<h[2-4][^>]*>|$)/gi;
```
dhananjay6561, line 98: minor: for the last question heading in a post, this answer capture runs to end of string, so it can swallow trailing wp markup like the author box or the more stories block that the post body strips before display. worst case is a noisy acceptedAnswer text, not invalid schema. could bound it to the article container or stop at a known footer marker.

dhananjay6561 (commented):
re review pass. no correctness bugs this time, ts is clean and all 66 unit tests pass. just a few low severity cleanups inline, nothing blocking.

dhananjay6561 (commented):
re review of the two new commits (db8ef2a, 5a84f11).

confirmed fixed:
- the orphaned imageobject doc block is back on getImageObjectSchema now, nice.
- occupationLocation dropped (schema wants a place there, not an employer) and worksFor already carries the affiliation, good call.

new hardening all reads correct and is covered by tests:
- toAbsoluteImageUrl for root relative and protocol relative image urls
- numeric/hex entity decode in stripTags with the safeFromCodePoint guard, so a junk entity like &#9999999999; cant crash the build
- c++, c#, csharp aliases (the regex already allowed + and #)
- tag index urls normalized to dashes + percent encoded, and they line up with the canonical the /tag/[slug] page emits (getStaticPaths builds from name, canonical uses the dash form)
- buildPageTitle now cuts at the last word boundary in the window instead of only when the space sits past char 40

tsc is clean and all 69 unit tests pass.

still open from the last pass (not touched): the per render content scans in community/[slug].tsx and technology/[slug].tsx (detectCodeLanguages / extractFaqs / countWords, countWords running twice) could use a useMemo or move into getStaticProps. low severity, not blocking. the faq tail capture note is minor too.

no new issues, looks good to me.

lib/structured-data.ts
```diff
@@ -139,6 +258,82 @@ export const getBreadcrumbListSchema = (items: BreadcrumbItem[]) => ({
...
+/**
+ * ItemList for a grid/collection of posts. Emitted standalone (has @context) so
+ * AI/search engines can map the ordered set of links on listing pages, which
+ * currently ship zero collection schema.
+ */
+export const getItemListSchema = (items: ListEntry[], listName?: string) => ({
+  "@context": "https://schema.org",
+  "@type": "ItemList",
+  ...(listName ? { name: listName } : {}),
+  itemListElement: toListItems(items),
```
dhananjay6561: 🟡 [important] the empty-list guard you just added to `getCollectionPageSchema` (94c77f1) doesn't cover this standalone builder. `authors/[slug].tsx:101` always builds `getItemListSchema(authoredItems, ...)` and always drops it into the `structuredData` array, so an author with 0 posts (or posts whose titles all get dropped by the new name/url filter) emits an `ItemList` with `itemListElement: []` — the same invalid empty ItemList you're clearing everywhere else.

note a call-site `authoredItems.length` check wouldn't fully cover it, since the filtering happens inside `toListItems` (2 entries with empty names still → `[]`). cleanest fix mirrors `getReviewSchema`: return `null` here when `toListItems(items)` is empty, and filter nulls out of the `structuredData` array on the author page.

lib/structured-data.ts
```diff
@@ -139,6 +258,82 @@ export const getBreadcrumbListSchema = (items: BreadcrumbItem[]) => ({
...
+};
+
+type ListEntry = { url: string; name: string; image?: string };
+
+const toListItems = (items: ListEntry[]) =>
+  items
+    // Drop entries without a usable name/url — a ListItem with an empty name is
+    // invalid schema (e.g. a WP post with a null title). Filtering before the
+    // map keeps `position` sequential (1, 2, 3, …).
+    .filter((it) => it.url && it.url.trim() && it.name && it.name.trim())
```
dhananjay6561, line 307: 🎉 [praise] nice detail running the filter *before* the map so `position` stays sequential (1, 2, 3…) instead of leaving gaps when an entry is dropped. easy to get wrong by filtering after the map.

lib/structured-data.ts
dhananjay6561: good catch, fixed in 8cfe264. `getItemListSchema` now returns null when nothing survives the name/url filter (same class of empty-ItemList guard as getCollectionPageSchema), and the authors/[slug] call site spreads it conditionally so a null never lands in the structuredData array. added a unit test pinning the null-on-empty behaviour too.

lib/structured-data.ts
dhananjay6561: thanks! yeah filtering before the map was deliberate so position stays 1,2,3 with no gaps.

lib/structured-data.ts
dhananjay6561: ✅ resolved in 8cfe264 — getItemListSchema now returns null when nothing survives the filter, and authors/[slug].tsx guards it with `...(authoredWorksSchema ? [authoredWorksSchema] : [])`. exactly the getReviewSchema pattern, so the all-filtered case is covered too. thanks.

dhananjay6561:
lgtm 🚀 schema validity is solid across the board now, every review point got addressed cleanly. the empty ItemList guard on getItemListSchema (8cfe264) mirrors getReviewSchema so the all filtered case is covered too, the a11y linkedin label is right, and the earlier rounds (occupationLocation, tag url normalization, word boundary title, entity decode safety, url less author fallback) all hold up. DCO green, E2E passing. nice work.

amaan-bhati:
# Claude Review Skill: Iteration 1

Reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) methodology (context gathering, high level, line by line, summary).

**Scope reviewed:** 29 files, +1376 / -423, 44 commits, base `main` @ `8cfe264`.

## 🚦 Verdict: 🔄 REQUEST CHANGES

| Severity | Count |
|---|---|
| 🔴 blocking | 1 |
| 🟡 important | 6 |
| 🟢 nit | 4 |
| 💡 suggestion | 2 |
| 🎉 praise | 4 |

**Verified locally:** `npx tsc --noEmit` clean, `npm run test:unit` 70/70 pass.

---

## 🔴 Blocking

**1. `datePublished` falls back to build time, which fabricates data and breaks hydration.**
`lib/structured-data.ts:409`

```ts
toISODate(datePublished) || toISODate(dateModified) || new Date().toISOString();
```

- A post with no usable WP date gets today's date as its publish date. That is fabricated structured data, the exact thing this PR refuses to do for `VideoObject` and `Rating`.
- `getBlogPostingSchema` is called in the component render body (`pages/community/[slug].tsx:167` area), so `new Date()` runs on both server and client. Server HTML and client hydration produce different JSON, giving a React hydration mismatch on the `ld+json` script.
- Also the value shifts on every ISR revalidate, so Google sees the publish date moving.

Suggested fix: drop `datePublished` when nothing valid exists, or fall back to a fixed value resolved in `getStaticProps`, not in render.

---

## 🟡 Important

**2. FAQPage is auto-extracted from any heading ending in `?`.**
`utils/contentSchema.ts:92`, pushed at `pages/community/[slug].tsx` and `pages/technology/[slug].tsx`

- The answer regex grabs everything up to the next `h2`-`h4`, so code blocks, tables and images get flattened into `Answer.text` and clipped at 900 chars. Mangled answers are what an AI engine will cite.
- The emitted `FAQPage` has no `url` or `@id` and sits alongside a `BlogPosting` on the same URL, so the page declares two page types.
- Google restricted FAQ rich results to gov/health sites in Aug 2023, so the upside here is small relative to the risk.

Suggested: gate on an explicit FAQ marker in the post, or drop `FAQPage` from article templates.

**3. `stripTags` decodes `&lt;` / `&gt;` on a JSON-LD path, dropping the repo's documented second layer.**
`utils/contentSchema.ts:79-80`

`utils/seo.ts` documents `decodeEntities` as script safe specifically because it leaves angle brackets encoded, and calls that a defence in depth layer behind `safeJsonLdStringify`. The new `stripTags` fully decodes them for FAQ answers. Not exploitable today since `safeJsonLdStringify` escapes at the boundary, but it silently removes the layer the module doc promises.

**4. Tag URL encoding is inconsistent between the two tag routes.**
`pages/tag/index.tsx:99` vs `pages/tag/[slug].tsx:32,36,49`

- Tag hub encodes: `/tag/${encodeURIComponent(...)}`
- Tag detail does not: breadcrumb, `CollectionPage.url` and `canonicalUrl` all use raw `tagSlug`

For a tag like `c#`, the hub emits `/tag/c%23` while the detail page emits `/tag/c#` where everything after `#` is a fragment. The `ItemList` entry and the canonical then disagree. Reuse the same encoding helper in both.

**5. Content extractors re-run on the client for no benefit.**
`pages/community/[slug].tsx:166-167,197`, `pages/technology/[slug].tsx:140-141,173`

`detectCodeLanguages`, `extractFaqs` and `countWords` all scan the full post HTML in the component body. That work already happened at build time and the output is already in the served HTML, so hydration repeats multiple regex passes over a large string purely to re-render identical `<script>` tags. Move to `getStaticProps` (the module doc already says these are meant for `getStaticProps`) or wrap in `useMemo`.

**6. `SoftwareApplication` claims the product is free.**
`lib/structured-data.ts:204-205`

```ts
isAccessibleForFree: true,
offers: { "@type": "Offer", price: "0", priceCurrency: "USD" },
```

Keploy ships paid tiers. This node is named `Keploy`, not `Keploy OSS`, so the claim reads as covering the product. Either scope the entity to the open source project or drop `offers` / `isAccessibleForFree`.

**7. `llms-full.txt` hardcodes metrics with no refresh path.**
`public/llms-full.txt`

`17,000+ GitHub stars`, `1,000,000+ VS Code installs`, `500+ articles`, `Gartner rated 4.6/5`. These go stale and there is nothing keeping them in sync. The Gartner line is also odd next to the PR's deliberate no-`AggregateRating` stance in JSON-LD.

---

## 🟢 Nits

**8.** `SearchResultsPage` schema is added to `pages/search.tsx:26` and `pages/community/search.tsx:64`, but both pages carry `<meta name="robots" content="noindex, follow">`. Google does not process structured data on noindexed pages, so this is dead markup.

**9.** `pages/index.tsx:88` sets `<title>{BLOG_TITLE}</title>` directly instead of `buildPageTitle(...)`. It happens to land at exactly 60 chars today, so any future edit silently breaks the invariant every other page enforces.

**10.** Four components deleted (`ReviewingAuthor`, `TagsPostPreview`, `latest-post`, `post-preview`, 352 lines). Confirmed unreferenced on `main`, so the deletion is safe, but it is unrelated to schema work and belongs in its own PR.

**11.** PR description says schema is wired into "404", but the PR removes the `BreadcrumbList` from `pages/404.tsx`. The removal is the right call, the description just does not match.

---

## 💡 Suggestions

**12.** `buildPageTitle` drops the ` | Keploy Blog` suffix for any title between 47 and 60 chars, which is most posts. That trades brand presence in SERP for a SEMrush warning that Google does not actually penalize. Worth confirming this is the intended trade.

**13.** The headline result (`1604 pages, 8656 blocks, 0 errors`) came from a local script that is not in this PR and not in CI. `test:unit` runs in `playwright.yml`, but nothing stops a future change from reintroducing invalid schema at scale. Committing the validator as a CI step would lock in the win.

---

## 🎉 Praise

- Empty `ItemList` and null-title `ListItem` guards in `toListItems` / `getItemListSchema` are the right shape, and the position resequencing after filtering is correct.
- Refusing to emit `VideoObject`, `HowToSection` and `AggregateRating` rather than fabricating fields is the correct call and well documented inline.
- The url-less fallback author node with the soft-404 reasoning is a genuinely subtle catch.
- The `AuthorDetailPage.spec.ts` fix (recursive `@type` walk plus `expect.poll` for the ISR fallback route) is a real improvement over the previous flaky assertion.

---

## Recommended path

Fix **#1** before merge. **#2**, **#4**, **#5**, **#6** are worth resolving in this PR. The rest can be follow ups.

dhananjay6561:
thanks for the thorough pass, this was genuinely useful. addressed everything actionable and pushed as signed micro commits. numbering below matches your review findings (not issue links):

blocking finding 1 is fixed. dropped the `new Date()` fallback so dates come from the post only and just get omitted when theres nothing valid. no more ld+json hydration mismatch and no more publish date drifting on every ISR revalidate.

the important ones:
- finding 2: pulled the auto FAQPage out of both article templates. the heading ending in "?" heuristic was flattening code/tables into the answer and declaring a second page type on the same url, not worth it. can re add behind an explicit in post marker later.
- finding 3: stripTags keeps angle brackets encoded now so the defence in depth layer stays.
- finding 4: tag slug is percent encoded in the breadcrumb, collectionpage url and canonical, matches the tag hub.
- finding 5: wrapped detectCodeLanguages/countWords in useMemo keyed on post content.
- finding 6: dropped isAccessibleForFree + offers from SoftwareApplication since keploy ships paid tiers.

quick nits too: finding 8 removed SearchResultsPage from the noindex search pages (kept the breadcrumb), finding 9 routed the home title through buildPageTitle.

left finding 7 (hardcoded llms metrics) as a content freshness follow up and finding 10 (dead component split) since its already in this prs history, noted 12 and 13.

all commits have DCO signoff, tsc is clean, unit suite is 71/71. updated the pr description with a round 1 table. mind taking another look when you get a sec?

amaan-bhati:

Re-review of the 10 commits pushed since Iteration 1 (`8cfe264..9963c21`). Delta reviewed: 11 files, +316 / -66.

**Verified locally on `9963c21`:** `npx tsc --noEmit` clean, `npm run test:unit` **74/74 pass**.

## 🚦 Verdict: 🔄 REQUEST CHANGES

Scoped: the round-1 fixes are good and the original schema work is ready. Everything below sits in the two things added *after* round 1, the FAQPage re-add and the new sitemap commit.

| Severity | Count |
|---|---|
| 🔴 blocking | 0 |
| 🟡 important | 5 |
| 🟢 nit | 5 |
| 💡 suggestion | 1 |
| 🎉 praise | 3 |

## ✅ Round-1 findings, verified

| # | Finding | Status |
|---|---|---|
| 1 🔴 | Fabricated `datePublished` | ✅ Fixed, dates omitted not invented |
| 2 🟡 | FAQPage from any `?` heading | ⚠️ Re-added behind a marker, see N1 |
| 3 🟡 | `stripTags` decoded angle brackets | ⚠️ Fixed as asked, caused N2 |
| 4 🟡 | Tag URL encoding mismatch | ✅ Fixed, all three sites encoded |
| 5 🟡 | Client re-scan of post HTML | ✅ Fixed via `useMemo` |
| 6 🟡 | Free/`offers` claim on the product | ✅ Fixed, with a test pinning it |
| 7 🟡 | Hardcoded `llms-full.txt` metrics | ⏭ Deferred, reasonable |
| 8 🟢 | `SearchResultsPage` on noindex pages | ✅ Fixed |
| 9 🟢 | Home `<title>` bypass | ✅ Fixed |
| 10 🟢 | Dead component split | ⏭ Declined, fair given history |
| 11 🟢 | Description drift | ✅ Fixed, then re-drifted, see N9 |

## 🟡 Important

**N1. The `<p>`-only answer rule silently drops list-based FAQs.**
`utils/contentSchema.ts` `extractFaqs`

Answers are built from `<p>` matches only. WordPress FAQ sections very often answer with `<ul>`/`<ol>`. Verified by running the extractor:

```
<h2>FAQ</h2>
<h3>What does Keploy record?</h3><ul><li>HTTP calls and their dependencies…</li></ul>
<h3>Does it need code changes?</h3><ul><li>No code changes are required…</li></ul>
→ extractFaqs() returns []
```

A post that opts in correctly gets nothing. Include `<li>` text (and keep excluding `<pre>`/`<table>`), which was the actual target of round-1 #2.

**N2. Answers now carry literal `&lt;` / `&gt;` into `Answer.text`.**
`utils/contentSchema.ts:79-82`

This one is on my round-1 #3, and the fix overshot. Verified:

```
<h3>How do I start recording traffic?</h3><p>Run &lt;keploy record&gt; in your terminal…</p>
→ answer: "Run &lt;keploy record&gt; in your terminal…"
```

An AI citing that answer reads the entity text, not `<keploy record>`. `safeJsonLdStringify` is already documented as the hard `</script>` guarantee and every injection site goes through it, so decoding angle brackets here is safe and gives correct content. Same issue already exists on the `description` path via `decodeEntities`, so worth settling the convention in one place rather than per module.

**N3. Sitemap `category` is not updated after `resolveLoc`, so category `lastmod` goes to the wrong page.**
`scripts/generate-sitemap.mjs` `buildEntries`

`loc` is resolved through the redirect map, but `latestByCategory.set(category, lastmod)` still uses the pre-redirect `category`. `next.config.js` has **13** literal `/community/X → /technology/X` redirects, so 13 posts that resolve to `/blog/technology/…` feed their `lastmod` into `/blog/community` instead. Derive the category from the resolved path.

**N4. The new redirect resolver has zero tests.**
`scripts/generate-sitemap.mjs`

The file was explicitly refactored (`invokedDirectly` guard plus `export { normalizePath, loadRedirectMap, resolveLoc, buildEntries }`) with a comment saying it exists so the resolver can be unit-tested, but no test was added. 37 redirects are now parsed and applied to production sitemap URLs on the strength of a manual check. Also note `test:unit` is globbed as `tests/lib/*.test.ts`, so a test needs to live there and import the `.mjs`.

**N5. The sitemap work is new scope introduced after review.**
`9963c21`

A8 was declared "🚧 Content/ops, not codeable here" in the PR description. A round-2 commit now adds a config-parsing redirect resolver to a PR already at +1641 / -438 across 30 files, in a file with its own GitHub Actions pipeline (`sync-sitemap.yml`, Google SC submit, IndexNow). It is a reasonable change but it should land on its own where it can be diffed against the live `sitemap.xml` before publishing. Recommend splitting `9963c21` out.

## 🟢 Nits

**N6.** `resolveLoc` is applied to post entries only. `staticEntries` (`/blog`, `/blog/community`, `/blog/technology`) skip it. None are in the redirect map today, so this is hardening only.

**N7.** `loadRedirectMap` has no guard for absolute `destination` values, `basePath: false`, or `has`/`missing` conditions. Confirmed zero of each exist today, so no live bug, but an absolute destination would produce `/blog/https:/…` and a conditional redirect would be applied unconditionally.

**N8.** When `resolveLoc` trips the 10-hop cap or the cycle guard it returns the last path silently, which is still a redirecting URL in the sitemap. Worth a `console.warn` so it shows up in the workflow log.

**N9.** Description re-drifted on FAQPage. It now says `FAQPage`/`Question`/`Answer` "were pulled in the round-1 review" and omits them from the emitted-types list, but `3f5d4ab` re-added them and `getFAQPageSchema(faqs, postUrl)` is live in both article templates. Also says 71/71 tests, actual is 74/74.

**N10.** `extractFaqs` boundary scan requires a matched `<hN>…</hN>` pair. Malformed WP markup with an unclosed heading means no boundary is found and the section runs to end of document, which degrades back toward the round-1 behaviour. Cheap to bound by a max section length.

## 💡 Suggestion

**N11.** `FAQPage` still sets `url` to the article URL. `@id: …#faq` plus `isPartOf` is a real improvement over round 1, but two nodes still claim the same `url`. Attaching the `Question` set to the article via `hasPart` would avoid the competing page-type entirely.

## 🎉 Praise

- The `datePublished` fix is the right one. Omitting the field with the reasoning written inline ("a missing datePublished is a soft recommended warning, whereas a fabricated one is exactly the invalid data this work refuses") is better than the workaround I suggested.
- Every fix carries the reasoning and a review reference in the comment, so the next reader gets the "why" without digging through PR history. The `SoftwareApplication` no-price comment is a good example.
- Marker-gated FAQ extraction with the level-aware section boundary is a much more careful design than the original heuristic, and the "same-level `?` heading stays a question" case is handled correctly.

## Recommended path

Fix **N1**, **N2**, **N3**. Split **N5** (`9963c21`) into its own PR, which also parks **N4**, **N6**, **N7**, **N8** there. **N9** is a description edit. Rest are optional.

dhananjay6561:
thanks for the iteration 2 pass, all genuinely useful. addressed everything in three signed commits (7d1826c, 98f83a3, aff95dc), 82/82 unit tests, tsc clean. decided to keep the sitemap work in this pr rather than split it, so i fixed n3/n4/n6/n7/n8 in place here.

**n1** — extractFaqs now builds answers from `<p>` AND `<li>` (drops `<pre>`/`<table>` first so code/tables still never leak). your bulleted-ul example returns 2 pairs now, added a test with it. 7d1826c

**n2** — decoding `&lt;`/`&gt;` now, so an answer reads `<keploy record>` not the entity. you're right that safeJsonLdStringify is the real `</script>` guarantee and every sink goes through it, so nothing's lost. test added with your exact example. 7d1826c

**n3** — real bug, fixed. added categoryFromLoc() and latestByCategory is keyed off the RESOLVED loc now, so a community post that 301s to technology lifts technology's lastmod. 98f83a3

**n4** — added tests/lib/generateSitemap.test.ts (imports the .mjs from tests/lib so the glob picks it up): normalizePath, resolveLoc chain + cycle, categoryFromLoc, and an n3 regression test with a moved post + a genuine community post so the buckets can't alias. 98f83a3

**n5** — kept it in this pr instead of splitting, so n6/n7/n8 landed here too.

**n6** — static entries run through resolveLoc now as well. 98f83a3

**n7** — loadRedirectMap skips has/missing (conditional) and absolute-url destinations, and honours basePath:false instead of prepending /blog. 98f83a3

**n8** — resolveLoc warns when it bails on the hop cap or a cycle (still a redirecting url), so it shows up in the workflow log. 98f83a3

**n9** — description fixed: FAQPage/Question/Answer are in the emitted list, the round-1 table row and AI4 row say re-enabled behind a marker, and the count is 82/82.

**n10** — bounded the FAQ section to a max length so unclosed heading markup can't sweep to end-of-doc. aff95dc

**n11** (suggestion) — leaving the FAQPage `@id` + isPartOf shape as-is for now, the hasPart-on-article rework felt like its own change; noted for a follow-up if we want to fully collapse the page-type.

amaan-bhati:

At `aff95dc`. Delta since Iteration 2: `7d1826c`, `98f83a3`, `aff95dc`.
**Verified:** tsc clean, `test:unit` **82/82 pass** (was 74).

| 🔴 0 | 🟡 2 | 🟢 2 | 💡 1 | 🎉 2 |
|---|---|---|---|---|

## ✅ Iteration 2 verified

| # | Status |
|---|---|
| N1 list answers dropped | ✅ `<p>` and `<li>` in document order, `<pre>`/`<table>` stripped first |
| N2 literal `&lt;`/`&gt;` | ✅ Decoded, with the `safeJsonLdStringify` reasoning stated |
| N3 wrong category `lastmod` | ✅ `categoryFromLoc` buckets by resolved path, plus a named regression test |
| N4 no resolver tests | ✅ 6 tests in `tests/lib/generateSitemap.test.ts` |
| N6 static entries unresolved | ✅ Now go through `resolveLoc` |
| N7 no has/absolute/basePath guards | ✅ All three skipped explicitly |
| N8 silent chain bail | ✅ Warns when the chain does not terminate |
| N9 description drift | ✅ FAQPage listed as re-enabled behind the marker |
| N10 unbounded FAQ section | ✅ Capped at 20k chars |

## 🟡 Important

**R1 (carried N5). Scope keeps growing.** Now +1808/-441 across 31 files. The sitemap redirect resolver is up to ~150 lines plus 100 lines of its own tests, inside a schema PR, in a file with its own Actions pipeline (`sync-sitemap.yml`, Google SC submit, IndexNow). The resolver is good work and well tested now, which makes it easier to split, not harder. Still recommend carving it out so it can be diffed against the live `sitemap.xml` before publishing.

**R2 (cross-PR, carried from #410 N3).** #411 emits `Review` server-side from `services/Tweets`; #410 makes the visible marquee `ssr: false`. Whichever merges second ships review markup with no visible reviews in the server HTML. Needs one decision across the two PRs, not two independent ones.

## 🟢 Nits

**R3.** `resolvedStaticEntries` resolves the `loc` but still keys `latestByCategory.get(category)` off the pre-redirect category, and there is no dedup against `postEntries`. If `/blog/community` ever gained a redirect you would emit a duplicate `<loc>` carrying the wrong category's `lastmod`. Hypothetical today, matching your own comment, but the same bug shape N3 just fixed one layer down.

**R4.** Nested `<li>` answers (`<li>a<ul><li>b</li></ul></li>`) end the outer lazy match at the inner `</li>`, so the outer tail after the nested list is dropped. Content still reads fine, just slightly truncated.

## 💡 Suggestion

**R5 (carried N11).** `FAQPage` still sets `url` to the article URL. `@id: …#faq` plus `isPartOf` is a real improvement, but two nodes still claim one URL. `hasPart` on the article would remove the competing page type.

## 🎉 Praise

- `categoryFromLoc` derives the bucket from the resolved loc rather than threading a second variable through, and the regression test names the finding it pins. That is the shape that stops a repeat.
- The three `add()` guards (has/missing, absolute destination, `basePath: false`) each say what would break without them, including the `/blog/https:/…` mangling. Cheap now, saves a confusing bug later.

dhananjay6561:
addressed iteration 3. pushed 22dbf81, 8a2975f, d3e330d.

r3 (static roots): fixed. the static archive-root entries resolved their loc through the redirect map but still keyed lastmod off the pre-redirect category and never deduped the resolved loc. so if /blog/community ever gained a redirect you'd emit a duplicate <loc> with the wrong category's lastmod, the same shape as n3 one layer up. added archiveRootCategory (kept categoryFromLoc scoped to post urls since a test pins that), bucket by the resolved category now, and dedup resolved static locs against each other and against post urls. covered by a named r3 regression test.

r4 (nested <li> truncation): left as is. the lazy match ends at the inner </li> so the outer tail drops, but regex genuinely cant balance nested lists and you already noted the content still reads fine. not worth a parser rewrite for a cosmetic trim.

r5 (faq url vs hasPart): left as is for now. the @id #faq + isPartOf already disambiguates and validates clean, and isPartOf/hasPart are just inverse ways to state the same part-whole. happy to switch to hasPart if you'd rather only one node claim the url.

r1 (carve out the sitemap resolver): keeping it in this pr. its tested and working, and splitting means a new branch + rebasing this on top + re-review for a modest gain. can carve it later if we want to diff sitemap.xml in isolation before publishing.

r2 (cross-pr with #410): this pr is internally consistent, testimonials render server-side so the Review markup matches whats visible. the mismatch only shows up once #410 makes the marquee ssr:false, so that pr should keep testimonials ssr or drop the markup. nothing to change here.

while in structured-data.ts also consolidated all json-ld into a single source of truth (context constant, shared org/blog @id reference helpers, Person/ProfilePage builders so the authors page stops hand-rolling them) and added three low-risk fields: isAccessibleForFree on articles, numberOfItems on ItemList/CollectionPage, and publisher+inLanguage on WebSite.

tsc clean, test:unit 87/87 (was 82).

dhananjay6561:
pushed an author avatar fix (95c33db, plus the fix commit before it).

the bug: ~22 author bylines were rendering a broken image. root cause was junk in wordpress, some authors have the literal strings "imag1" / "image" (or empty) in ppmaAuthorImage instead of a real photo. that value went straight into next/image, so /_next/image?url=imag1 returned a 400 and the byline avatar broke on every post those authors wrote. the authors index hid it because it had a brittle hardcoded check for exactly "imag1"/"image", so the same author looked fine there but broken on their post.

fix: added resolveAuthorAvatar() + AUTHOR_AVATAR_PLACEHOLDER in lib/constants. anything thats not a real http(s) url or a root-relative path collapses to the placeholder, a genuine url passes through untouched. routed every avatar render path through it, the byline (community/technology [slug]), authors index (AuthorMapping), author hero (AuthorHero) and the bottom author card (AuthorCard). also stopped the junk leaking into the json-ld author.image, schema now gets a real url or nothing.

verified on a real build: kanishak chaurasia and keploy team posts (both had junk) now show the clean placeholder at every spot, zero url=imag1 on the page, placeholder optimizes 200. covered by tests/lib/resolveAuthorAvatar.test.ts.

the real content fix (uploading real photos for those 22, or clearing the junk field) is handed to the seo team for wordpress, this just makes the site resilient so bad data can never render a broken avatar again.

tsc clean, test:unit 91/91.

amaan-bhati:

At `95c33db` (6 new commits). Reviewed from the diff. **`tsc` / `test:unit` not run this round** (no disk space on the review machine), so treat CI as the source of truth for those.

## 🚦 Verdict: 🔄 REQUEST CHANGES

| 🔴 blocking | 🟡 important | 🟢 nit | 💡 suggestion | 🎉 praise |
|---|---|---|---|---|
| 1 | 1 | 0 | 1 | 3 |

## ✅ Iteration 3 verified

| # | Status |
|---|---|
| R2 cross-PR `Review` vs `ssr: false` | ✅ Dropped from the homepage, with the reasoning and the #410 reference written inline. `getReviewSchema` kept and tested so it can be re-wired if the wall returns to SSR. Correct call |
| R3 static roots: wrong category, no dedup | ✅ `archiveRootCategory` buckets by the resolved root, and dedup now runs against both `staticSeen` and the post `seen` set |
| R5 `FAQPage` competing page type | ⏭ Still open as a suggestion, unchanged |

Also resolved from earlier rounds: the hardcoded `llms-full.txt` metrics (17k stars, 1M installs, Gartner 4.6/5, 500+ articles) are gone, which closes Iteration 1 finding 7.

## 🔴 Blocking

**F1. `lib/constants.ts` conflicts with #410, and the second merge breaks the build.** Same finding as on #410, needs fixing once.

| | #411 (this PR) | #410 |
|---|---|---|
| `thumbnil.png` import | still imported | removed, file deleted |
| `AUTHOR_AVATAR_PLACEHOLDER` | `/blog/images/author.png` | `/blog/images/author.webp` |
| `author.png` in tree | present | deleted |

If #410 lands first, this branch reintroduces an import for a deleted file (module-not-found at build) and `AUTHOR_AVATAR_PLACEHOLDER` points at a deleted `author.png`. That is precisely the broken-avatar failure `resolveAuthorAvatar` was written to prevent, reintroduced through the placeholder path, and `resolveAuthorAvatar` will pass it through happily since it only checks that the value is root-relative. Neither branch's CI catches it alone.

## 🟡 Important

**R1 (carried). Scope.** Unchanged from Iteration 3, and this round added a JSON-LD builder refactor, an author-avatar fix, a sitemap change and an `llms-full.txt` rewrite on top. The sitemap resolver in particular is now well tested, which makes splitting it cheaper, not dearer.

## 💡 Suggestion

**R5 (carried).** `FAQPage` still sets `url` to the article URL. `hasPart` on the article would drop the competing page type.

## 🎉 Praise

- `isAccessibleForFree` on the article with the contrast spelled out against the `SoftwareApplication` builder that deliberately omits it. That is the distinction Iteration 1 finding 6 was actually about, and getting it right in one place while explaining the other is better than either change alone.
- `orgReference()` / `blogReference()` / `SCHEMA_CONTEXT` do the thing `@id` de-fragmentation needs: byte-identical references, with a comment telling the next author to build from these rather than hand-rolling. That is what keeps the entity graph from drifting back apart.
- `resolveAuthorAvatar` is a real bug caught outside the review loop, with the concrete failure named (`ppmaAuthorImage` of `"imag1"` becoming `/_next/image?url=imag1` and a 400 on every post by that author) and a test alongside it.

dhananjay6561:
f1 (the cross pr constants + asset conflict): resolved from this side in 1a38e61. aligned this branch to #410's asset decisions so neither merge order breaks the build:

- dropped the dead thumbnil.png import (kills the module not found half)
- AUTHOR_AVATAR_PLACEHOLDER and every hardcoded ref now point at /blog/images/author.webp (constants, the structured data placeholder guard, the community + technology reviewer/author fallbacks, and the resolveAuthorAvatar test)
- added public/images/author.webp (byte identical to #410) and removed author.png + thumbnil.png

so both branches now agree: webp placeholder, no author.png, no thumbnil import. verified with git merge tree that the two landmines you flagged (an import of a deleted file, and a placeholder pointing at a deleted asset) are gone in both merge orders. the branches still have ordinary textual conflicts in the files they both touch (author components, constants, structured data), but those resolve by hand at merge now with no silent build break.

tsc clean, test:unit 91/91, lint clean. r1 (scope) and r5 (faq hasPart) still noted, leaving as is for this pr.

components/latest-post.tsx
amaan-bhati, line 1: Add a pointer or comment in the pr description for why you have deleted these 4 files, good to keep things written in the pr description for context of anyone seeing it in future.

amaan-bhati (approved):
## Approving. Verified locally at `1a38e61`

Nothing failing: CI green on this head (build, lighthouse, E2E chromium, DCO), `mergeable: MERGEABLE`. Locally `npm run test:unit` **91/91**, `npx tsc --noEmit` clean.

Went past the builders and checked the real output.

**Sitemap, generated against live WordPress:**

```
Generated public/sitemap.xml with 525 WordPress posts as input.

total <loc>: 521          duplicate <loc>: 0
non-https: 0              with #fragment: 0        double slashes: 0
lastmod: 521, malformed: 0
post locs: 518  (community 481 / technology 37)
non-post locs: 3  ->  /blog, /blog/community, /blog/technology
```

Zero duplicate `<loc>` on 525 real posts is the R3 dedup holding outside fixtures.

**JSON-LD actually emitted, three page shapes.** Every block parses, `@context` is `https://schema.org` uniformly, `@id` graph stable across pages, no dangling references:

```
/blog             WebSite @id=…/blog/#website · CollectionPage · Organization
                  @id=…/#organization · Blog @id=…/blog/#blog · SoftwareApplication
/blog/technology  BreadcrumbList · CollectionPage · + the same @id nodes
/blog/community/… BreadcrumbList · BlogPosting · FAQPage · SoftwareSourceCode

numberOfItems accurate:   /blog 6=6      /blog/technology 22=22
BlogPosting:              9/9 fields present, headline 53 chars
                          isAccessibleForFree=True (SoftwareApplication still omits it)
FAQPage:                  @id=…#faq, 10 questions, 0 missing an answer
```

`R2` confirmed: no `Review` node anywhere in the emitted output.

## 💡 Two follow-ups, neither blocking

- **R5, smaller than before.** FAQPage `@id` now carries `#faq`, but its `url` is still the bare article URL, so it and the BlogPosting both present as the page at that URL. Dropping `url` (the `#faq` `@id` already identifies it) closes it.
- **Sitemap drop accounting.** The script logs `525 posts as input` and emits 518 post locs, but never logs the output count or why any were dropped. The 7 are almost certainly uncategorised posts, correct by design, but that is my inference and I could not tell from the output. One line reporting `emitted N, skipped M, collapsed K` would make a future 500 to 400 regression obvious, and this feeds the Search Console / IndexNow pipeline.

Approving.

components/latest-post.tsx
dhananjay6561: done, added a "removed files why" section to the pr description with the reasoning. short version: all four were dead components with zero importers on main (verified by grep), leftovers from an older layout the current templates already replaced. reviewingauthor also pointed at the old author.png and was part of the review surface we deliberately stopped emitting. the two unused images (author.png replaced by author.webp, and thumbnil.png) went with them. full context is written up in the description now for anyone reading later.

nehagup (commented):
## Review — harden Article schema + site health

**Recommendation: Approve with nits.** Careful schema-hardening with unusually good unit tests. The changes that actually prevent Rich-Results rejections / manual-action risk are correct: **dates resolved from post values only and omitted (never fabricated) when absent** (fixes a real hydration/ISR-drift bug), **no fabricated `Rating`/`AggregateRating`**, `image` always emitted as an absolute-URL `ImageObject`, and descriptions tag-stripped/entity-decoded through `safeJsonLdStringify` (no `</script>` injection). Per-post author `@id` and the ProfilePage Person `@id` derive from the same slug so they merge; the url-less fallback author correctly carries no `@id`.

Schema is computed in the page data layer and the sitemap runs as a build script — neither adds runtime functions. The one caveat: it touched REVALIDATE_* constants. If any ISR revalidate interval was lowered, pages regenerate more often = more function invocations.


### Should-fix
- **`lib/structured-data.ts` — `headline` has no length cap.** `getBlogPostingSchema` sets `headline: title` verbatim; Google recommends ≤110 chars and ignores Article rich results for over-long headlines. The PR caps the `<title>` but not `headline`, and 113 posts were flagged long-title. Add a ≤110 truncation (or verify on Rich Results Test with a long-title post).
- **`scripts/generate-sitemap.mjs` (`loadRedirectMap`)** — `next.config.js` throws at load if `WORDPRESS_API_URL` is unset; the dynamic import is in a try/catch that logs and continues, so a missing env var **silently** yields zero folded redirects → the sitemap can emit `<loc>`s that 301 (the exact issue this is meant to fix). Surface a louder failure or assert the redirect map is non-empty in the real build path.

### Nits
- `pages/authors/[slug].tsx:36` — an `import` sits after a function decl (legal, hoisted, but out of place).
- `getSearchResultsPageSchema` / `getReviewSchema` are now dead exports (intentional, kept for re-wiring) — worth a comment so a future reader doesn't assume they're live.
- Fallback author name differs: `"Keploy Team"` (post schema) vs `"Keploy Author"` (`authors/[slug].tsx:47`) — cosmetic.
- `speakableSelectors` is valid but Google's Speakable is news-only — aspirational, not a ranking lever.

### Before merge (build safety)
- Grep-confirm the **4 deleted components** (`ReviewingAuthor`, `TagsPostPreview`, `latest-post`, `post-preview`) and **2 deleted images** (`author.png`) have no remaining importers/references on `main` — a stray import would 404/break the build.

### Verify after deploy
Google Rich Results Test (long-title TechArticle, null-`featuredImage` post, an FAQ post, an author ProfilePage) and GSC Enhancements for new warnings; confirm sitemap `<loc>`s resolve to 200s (no 301s).

scripts/generate-sitemap.mjs
```diff
@@ -152,7 +155,102 @@ async function fetchAllPosts(endpoint) {
...
+
+  // vercel.json — sources already include the /blog basePath.
+  try {
+    const vercel = JSON.parse(
+      await readFile(path.join(repoRoot, "vercel.json"), "utf8")
+    );
+    for (const r of vercel.redirects ?? []) add(r);
+  } catch (error) {
+    console.warn("[generate-sitemap] Could not read vercel.json redirects:", error.message);
+  }
```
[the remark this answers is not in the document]
dhananjay6561: good catch, agreed. the warn meant an incomplete redirect map could ship silently, and since sync-sitemap.yml auto commits sitemap.xml to main, that bad sitemap would go out with nobody looking. changed both config loads (vercel.json and next.config.js) to throw so the action fails and the last good sitemap stays instead. left the resolveLoc chain warning (line 226) as is though, thats a best effort emit on a hop cap or cycle, not a config load failure, so throwing there would just hard fail the sitemap on a redirect loop which is worse. done in ba09062.

pages/tag/[slug].tsx
```diff
@@ -7,33 +7,64 @@ import Container from "../../components/container";
...
 export default function PostByTags({ postsByTags, preview, tagSlug: tagSlugProp }) {
   const posts = postsByTags?.edges || [];
   const router = useRouter();
   const tagSlug = tagSlugProp || (Array.isArray(router.query.slug) ? router.query.slug[0] : (router.query.slug || ''));
+  // Percent-encode the slug for every emitted URL — the tag hub already does
+  // (pages/tag/index.tsx), so without this a tag like "c#" produces /tag/c%23
+  // on the hub but a raw /tag/c# here, where everything after # is a fragment.
+  // The ItemList entry and the canonical would then disagree.
+  const encodedTagSlug = encodeURIComponent(tagSlug);
   const tagDisplay = tagSlug.replace(/-/g, ' ').replace(/\b\w/g, c => c.toUpperCase()) || 'All Topics';
```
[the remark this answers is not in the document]
Comment (dhananjay6561) [reply pages/tag/[slug].tsx:22]: i think this one is a false positive. the page is fallback blocking, so getStaticProps always runs server side before any html gets emitted, and it already normalizes the slug there (slug.replace(/\s+/g, "-")) and passes that normalized value as the tagSlug prop. the component reads tagSlugProp first, so encodedTagSlug is built from the normalized slug, the exact same string used to fetch the posts. the router.query branch only kicks in if the prop is falsy, which never happens on a served page. so hitting /tag/some%20tag gives a prop thats already normalized, and canonical, json ld and the fetch all agree. leaving as is, lmk if im missing something.

nehagup (approved):

lib/structured-data.ts
```diff
@@ -251,18 +534,261 @@ export const getBlogPostingSchema = ({
...
+    hasOccupation: {
+      "@type": "Occupation",
+      name: "Technical Author",
+      // No occupationLocation: schema.org expects a geographic Place there, not
+      // an employer. The Keploy affiliation is already modeled via worksFor.
+    },
+    worksFor: orgReference(),
+  };
+  if (knowsAbout && knowsAbout.length) node.knowsAbout = knowsAbout;
+  if (image) node.image = image;
```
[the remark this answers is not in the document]
nehagup: @dhananjay6561 please fix.

nehagup (commented):
## Follow-up: tested the deleted files — safe to merge, one cleanup

Because this PR removes several components + images, I checked out the head branch and grepped the whole tree for every deletion. **The deletions are safe** — `build` and `E2E (chromium)` are green at head, and there are no live references:

| Deleted file | Status | Evidence |
|---|---|---|
| `components/ReviewingAuthor.tsx` | ✅ safe | 0 references |
| `components/TagsPostPreview.tsx` | ✅ safe | 0 references |
| `components/latest-post.tsx` | ✅ safe | the only "hit" is a log string in `pages/404.tsx` ("latest-posts rails empty"), not the component |
| `components/post-preview.tsx` | ✅ safe | 0 references |
| `public/images/thumbnil.png` | ✅ safe | the `thumbnil` references are a **different remote** URL (`wp.keploy.io/.../thumbnil-.png`), not this local file |
| `public/images/author.png` | ⚠️ see below | migrated to `author.webp` everywhere live; one surviving ref is in dead code |

### Should-fix (non-blocking): delete `components/author-description.jsx` too
The avatar placeholder was correctly migrated `author.png → author.webp` across every live path — `lib/constants.ts` (`AUTHOR_AVATAR_PLACEHOLDER`), `lib/structured-data.ts`, both `pages/technology/[slug].tsx` and `pages/community/[slug].tsx`, plus a `resolveAuthorAvatar` test.

The one place it was **missed** is `components/author-description.jsx:123`, which still hardcodes `src={"/blog/images/author.png"}` (the deleted file) in its `avatarImgSrc === "n/a"` fallback. This doesn't 404 today only because `author-description.jsx` appears to be **orphaned** — nothing imports `AuthorDescription`, and the `#author-description-placeholder` div in `post-body.tsx` is never populated. So it's dead code that references a now-deleted asset.

Recommendation: **delete `components/author-description.jsx`** alongside the other four orphan components this PR already removes (it's the last thing pointing at `author.png`). If there's any reason to keep it, at minimum change line 123 `author.png → author.webp`.

Net: the critical-file deletions are safe to merge; this is a cleanup, not a blocker.

<sub>🤖 Assisted review via Claude Code (verified against head `ai-cite-health`).</sub>

PR #414, opened 2026-08-15
feat(blog): notify google chat on newsletter form submit

dhananjay6561:
Forwards every blog newsletter/lead form submission to a Google Chat space, so the team gets pinged in real time when a new lead comes in. Same approach we already use for the request-a-trial form on the landing repo.

## Where the form data goes today

The newsletter form (under the ad banner on every blog post) already sends data to two places, both MongoDB. This PR does **not** change either of them:

- **Newsletter subscription** → api-server, via the GraphQL `subscription(guestInput)` mutation, lands in the `subscriptions` collection (upsert by email, so repeat emails get merged). This is the actual subscriber list.
- **Lead / MQL copy** → telemetry `POST /blog-mql`, verified server-side via reCAPTCHA Enterprise, lands in the `keploy-telemetry.blog_mql` collection (one row per submit, for lead tracking).

So the data is stored, but nobody gets notified when a new lead arrives. That's the gap this PR fills.

## What changed

- New server-side route `pages/api/blog-lead-notify.ts` that holds the webhook secret and posts a formatted message to the space.
- `subscribe-newsletter.tsx` calls it fire-and-forget, alongside the existing subscription + blog-mql capture (does not touch or gate either).
- Documented the env var in `.env.local.example`.

## Action needed to turn it on

- Add a **server-only** env var `GOOGLE_CHAT_WEBHOOK_URL` in the blog-website Vercel project (not `NEXT_PUBLIC_*`, so it never reaches the browser).
- Create the webhook in the Chat space: **Apps & integrations → Webhooks → add → copy URL**.
- Until this env is set, everything keeps working exactly as before, just no Chat message is delivered.

## Notes

- **Fail-open:** if the var is unset or the POST fails, the user's submit and the newsletter subscription are unaffected.
- **Delivery-safe & consistent with repo conventions:** aborts the Chat fetch at 15s and caps field lengths (name 120 / email 254 / company 160 / page 500), mirroring `blog-mql.ts`; logs gated on `NODE_ENV`.
- **Basic bot filter:** a hidden honeypot field drops autofill bots; server-side email validation.
- **Not gated with auth (deliberate follow-up):** this mirrors landing's `/trial/submit`, which is also a public unauthenticated POST. Stronger gating via reCAPTCHA Enterprise (forwarding + verifying the token the client already mints) needs the GCP project-id + api-key secrets that live in telemetry, not this repo — tracked as a follow-up rather than half-gated here.

### DIFF
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -49,6 +49,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
   // Honeypot — hidden from humans; bots that auto-fill every field trip it.
   const [companyWebsite, setCompanyWebsite] = useState('');
   const [subscribed, setSubscribed] = useState<boolean>(false);
+  // In-flight guard: blocks repeat clicks so we don't fire duplicate lead /
+  // Chat-notify POSTs (the subscription upserts by email, but the Chat space
+  // gets one ping per click otherwise).
+  const [submitting, setSubmitting] = useState<boolean>(false);
   const [emailError, setEmailError] = useState('');
   const message = "NEWSLETTER"
   // Don't load Google's script for every reader — only once someone actually
@@ -93,6 +97,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
     return 
   }
 
+    // Guard against double-submits — one Chat ping per click otherwise.
+    if (submitting) return;
+    setSubmitting(true);
+
     const payload = {
       fullName,
       email,
@@ -133,10 +141,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
       }),
     }).catch(() => {});
 
-    handleSubscribe(payload)
+    handleSubscribe(payload).finally(() => setSubmitting(false));
   };
   const isSubscribeDisabled = ()=>{
-    return Boolean(!email || !fullName || !companyName)    
+    return Boolean(!email || !fullName || !companyName || submitting)
   }
   return (
     <div className="flex flex-col">
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -80,6 +80,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
   const submitHandler = (e) => {
     e.preventDefault();
 
+    // Honeypot — bots fill the hidden company_website field; humans don't.
+    // Drop silently client-side too (the endpoint re-checks server-side).
+    if (companyWebsite) return;
+
       if (!isValidEmail(email)) {
     setEmailError("Please enter a valid email address."); 
 
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -128,6 +128,7 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
         email: email.trim().toLowerCase(),
         companyName: companyName.trim(),
         company_website: companyWebsite,
+        source: 'blog-newsletter',
         page,
       }),
     }).catch(() => {});
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -51,7 +51,11 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
   const [subscribed, setSubscribed] = useState<boolean>(false);
   // In-flight guard: blocks repeat clicks so we don't fire duplicate lead /
   // Chat-notify POSTs (the subscription upserts by email, but the Chat space
-  // gets one ping per click otherwise).
+  // gets one ping per click otherwise). The ref is the airtight guard — a fast
+  // double-click fires two handlers before React re-renders, so reading the
+  // `submitting` state (or the disabled button) still sees the stale `false`.
+  // The state drives the disabled/opacity UI; the ref decides who actually runs.
+  const submittingRef = useRef<boolean>(false);
   const [submitting, setSubmitting] = useState<boolean>(false);
   const [emailError, setEmailError] = useState('');
   const message = "NEWSLETTER"
@@ -98,7 +102,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
   }
 
     // Guard against double-submits — one Chat ping per click otherwise.
-    if (submitting) return;
+    // Check/set the ref synchronously so a second click in the same tick can't
+    // slip through before the state re-renders.
+    if (submittingRef.current) return;
+    submittingRef.current = true;
     setSubmitting(true);
 
     const payload = {
@@ -141,7 +148,10 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
       }),
     }).catch(() => {});
 
-    handleSubscribe(payload).finally(() => setSubmitting(false));
+    handleSubscribe(payload).finally(() => {
+      submittingRef.current = false;
+      setSubmitting(false);
+    });
   };
   const isSubscribeDisabled = ()=>{
     return Boolean(!email || !fullName || !companyName || submitting)
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -1,4 +1,5 @@
 import { useRef, useState } from "react";
+import { useRouter } from "next/router";
 import styles from "./subscribe-newsletter.module.css";
 import { newsLetterSubscriptionUrl } from '../services/constants'
 import { useInvisibleRecaptcha, RecaptchaAttribution } from "../lib/use-invisible-recaptcha";
@@ -39,6 +40,7 @@ export const subscribeMutation = (formData: { fullName: string, email: string, c
 
 
 export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) {
+  const router = useRouter();
   const myComponent = useRef<HTMLDivElement>(null);
   const [isVisible, setVisible] = useState<boolean>(true);
   const [fullName, setFullName] = useState('');
@@ -107,6 +109,21 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
         recaptchaToken,
       });
     });
+
+    // Google Chat notification — fire-and-forget, fail-open. Forwards the lead
+    // to a Chat space via the server-side /api/blog-lead-notify handler (which
+    // holds the webhook secret). Never gates the subscription.
+    fetch(`${router.basePath || ''}/api/blog-lead-notify`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      keepalive: true,
+      body: JSON.stringify({
+        fullName: fullName.trim(),
+        email: email.trim().toLowerCase(),
+        companyName: companyName.trim(),
+        page,
+      }),
+    }).catch(() => {});
 
     handleSubscribe(payload)
   };
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -46,6 +46,8 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
   const [fullName, setFullName] = useState('');
   const [email, setEmail] = useState('');
   const [companyName, setCompanyName] = useState('');
+  // Honeypot — hidden from humans; bots that auto-fill every field trip it.
+  const [companyWebsite, setCompanyWebsite] = useState('');
   const [subscribed, setSubscribed] = useState<boolean>(false);
   const [emailError, setEmailError] = useState('');
   const message = "NEWSLETTER"
@@ -121,6 +123,7 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
         fullName: fullName.trim(),
         email: email.trim().toLowerCase(),
         companyName: companyName.trim(),
+        company_website: companyWebsite,
         page,
       }),
     }).catch(() => {});
@@ -146,6 +149,18 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean }) 
               onSubmit={submitHandler}
               onFocusCapture={() => setCaptchaActive(true)}
             >
+              {/* Honeypot — hidden from humans (off-screen, not tabbable, no
+                  autofill); a filled value marks the submit as a bot. */}
+              <input
+                type="text"
+                name="company_website"
+                tabIndex={-1}
+                autoComplete="off"
+                aria-hidden="true"
+                value={companyWebsite}
+                onChange={(e) => setCompanyWebsite(e.target.value)}
+                style={{ position: "absolute", left: "-9999px", width: 1, height: 1, opacity: 0 }}
+              />
               <input
                 type="text"
                 className="rounded px-4 py-2 border border-gray-300 focus:outline-none focus:ring-2 focus:ring-primary-500 focus:border-transparent"
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -12,72 +12,10 @@ import type { NextApiRequest, NextApiResponse } from "next";
 //     the webhook URL is never exposed to the client, git, or logs.
 //   • Create it in Google Chat: open the space → Apps & integrations → Webhooks
 //     → add one → copy the URL into the env var. Rotate by deleting/recreating.
-//   • Abuse: a hidden honeypot field + a best-effort per-IP rate limit keep
-//     casual flooding out. Stronger gating (reCAPTCHA Enterprise verification,
-//     matching the /blog-mql path) is tracked as a follow-up.
 // If the env var is unset the submission still succeeds for the user, but the
 // lead is NOT delivered.
 
 const EMAIL_RE = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
-
-// Best-effort in-memory per-IP rate limit. On serverless this is per-instance,
-// not global, so it caps bursts against a hot instance rather than guaranteeing
-// a hard ceiling — enough to blunt casual flooding of the chat space without
-// pulling in an external store.
-const RATE_LIMIT_MAX = 5; // requests
-const RATE_LIMIT_WINDOW_MS = 60_000; // per minute, per IP
-const RATE_MAP_MAX_KEYS = 10_000; // evict stale IPs past this so the map can't grow unbounded
-const rateHits = new Map<string, number[]>();
-
-function isRateLimited(ip: string): boolean {
-  const now = Date.now();
-  const cutoff = now - RATE_LIMIT_WINDOW_MS;
-
-  // Opportunistically drop IPs whose most recent hit is outside the window, so
-  // the map doesn't accumulate an entry per distinct IP for the instance's life.
-  if (rateHits.size > RATE_MAP_MAX_KEYS) {
-    rateHits.forEach((times, key) => {
-      if (times.length === 0 || times[times.length - 1] <= cutoff) rateHits.delete(key);
-    });
-  }
-
-  const hits = (rateHits.get(ip) || []).filter((t) => t > cutoff);
-  hits.push(now);
-  rateHits.set(ip, hits);
-  return hits.length > RATE_LIMIT_MAX;
-}
-
-// Identify the client for rate limiting. On Vercel the trustworthy client IP is
-// `x-real-ip` (set by the platform). Do NOT use the leftmost x-forwarded-for
-// token — that end is client-supplied and lets an attacker rotate it to dodge
-// the limit; the real IP is the LAST hop the trusted proxy appends.
-function clientIp(req: NextApiRequest): string {
-  const realIp = req.headers["x-real-ip"];
-  if (typeof realIp === "string" && realIp.trim()) return realIp.trim();
-
-  const fwd = req.headers["x-forwarded-for"];
-  const raw = Array.isArray(fwd) ? fwd[0] : fwd;
-  if (raw) {
-    const parts = raw.split(",").map((s) => s.trim()).filter(Boolean);
-    if (parts.length) return parts[parts.length - 1];
-  }
-  return req.socket.remoteAddress || "unknown";
-}
-
-// Strip characters that carry meaning in Google Chat `text` messages so user
-// input can't inject formatting (*bold* / _italic_), fake clickable links
-// (<url|label>), or break the layout with newlines. Also caps length.
-function sanitize(value: string, max = 200): string {
-  return value
-    .replace(/[<>|*_`\r\n]/g, " ")
-    .replace(/\s+/g, " ")
-    .trim()
-    .slice(0, max);
-}
-
-// Track the "webhook not configured" warning so it's logged once per instance
-// instead of on every submit — the off state is expected until the env is set.
-let warnedMissingWebhook = false;
 
 export default async function handler(
   req: NextApiRequest,
@@ -93,14 +31,9 @@ export default async function handler(
     unknown
   >;
 
-  // Honeypot — the form renders a hidden `company_website` input that humans
-  // never see; a filled value means a bot. Silently accept and drop.
+  // Honeypot — silently accept (and drop) bot submissions.
   if (data.company_website) {
     return res.status(200).json({ ok: true, delivered: false });
-  }
-
-  if (isRateLimited(clientIp(req))) {
-    return res.status(429).json({ ok: false, error: "rate_limited" });
   }
 
   const name = String(data.fullName ?? "").trim();
@@ -112,29 +45,24 @@ export default async function handler(
   }
 
   const lead = {
-    name: sanitize(name, 120),
-    email: sanitize(email, 254),
-    company: sanitize(String(data.companyName ?? ""), 160),
-    page: sanitize(String(data.page ?? ""), 500),
+    name,
+    email,
+    company: String(data.companyName ?? "").trim(),
+    page: String(data.page ?? ""),
     submittedAt: new Date().toISOString(),
   };
 
   const webhook = process.env.GOOGLE_CHAT_WEBHOOK_URL;
   if (!webhook) {
     // No PII in logs. Set GOOGLE_CHAT_WEBHOOK_URL to deliver leads to the space.
-    // Expected steady state until the env is set, so warn once, not every submit.
-    if (!warnedMissingWebhook) {
-      warnedMissingWebhook = true;
-      console.warn(
-        "[blog-lead] GOOGLE_CHAT_WEBHOOK_URL is not configured — leads accepted but NOT delivered. Set the env var to enable delivery.",
-      );
-    }
+    console.error(
+      "[blog-lead] GOOGLE_CHAT_WEBHOOK_URL is not configured — lead accepted but NOT delivered. Set the env var to enable delivery.",
+    );
     return res.status(200).json({ ok: true, delivered: false });
   }
 
   try {
-    // page is rendered as plain (sanitized) text, not a <url|label> link, so a
-    // direct POST can't inject a misleading clickable URL.
+    // Google Chat renders *bold* / _italic_ and <url|label> in `text` messages.
     const text =
       `*📨 New Keploy blog subscriber*\n` +
       `*Name:* ${lead.name}\n` +
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -26,21 +26,42 @@ const EMAIL_RE = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
 // pulling in an external store.
 const RATE_LIMIT_MAX = 5; // requests
 const RATE_LIMIT_WINDOW_MS = 60_000; // per minute, per IP
+const RATE_MAP_MAX_KEYS = 10_000; // evict stale IPs past this so the map can't grow unbounded
 const rateHits = new Map<string, number[]>();
 
 function isRateLimited(ip: string): boolean {
   const now = Date.now();
   const cutoff = now - RATE_LIMIT_WINDOW_MS;
+
+  // Opportunistically drop IPs whose most recent hit is outside the window, so
+  // the map doesn't accumulate an entry per distinct IP for the instance's life.
+  if (rateHits.size > RATE_MAP_MAX_KEYS) {
+    rateHits.forEach((times, key) => {
+      if (times.length === 0 || times[times.length - 1] <= cutoff) rateHits.delete(key);
+    });
+  }
+
   const hits = (rateHits.get(ip) || []).filter((t) => t > cutoff);
   hits.push(now);
   rateHits.set(ip, hits);
   return hits.length > RATE_LIMIT_MAX;
 }
 
+// Identify the client for rate limiting. On Vercel the trustworthy client IP is
+// `x-real-ip` (set by the platform). Do NOT use the leftmost x-forwarded-for
+// token — that end is client-supplied and lets an attacker rotate it to dodge
+// the limit; the real IP is the LAST hop the trusted proxy appends.
 function clientIp(req: NextApiRequest): string {
+  const realIp = req.headers["x-real-ip"];
+  if (typeof realIp === "string" && realIp.trim()) return realIp.trim();
+
   const fwd = req.headers["x-forwarded-for"];
   const raw = Array.isArray(fwd) ? fwd[0] : fwd;
-  return (raw?.split(",")[0].trim() || req.socket.remoteAddress || "unknown");
+  if (raw) {
+    const parts = raw.split(",").map((s) => s.trim()).filter(Boolean);
+    if (parts.length) return parts[parts.length - 1];
+  }
+  return req.socket.remoteAddress || "unknown";
 }
 
 // Strip characters that carry meaning in Google Chat `text` messages so user
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -12,10 +12,51 @@ import type { NextApiRequest, NextApiResponse } from "next";
 //     the webhook URL is never exposed to the client, git, or logs.
 //   • Create it in Google Chat: open the space → Apps & integrations → Webhooks
 //     → add one → copy the URL into the env var. Rotate by deleting/recreating.
+//   • Abuse: a hidden honeypot field + a best-effort per-IP rate limit keep
+//     casual flooding out. Stronger gating (reCAPTCHA Enterprise verification,
+//     matching the /blog-mql path) is tracked as a follow-up.
 // If the env var is unset the submission still succeeds for the user, but the
 // lead is NOT delivered.
 
 const EMAIL_RE = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
+
+// Best-effort in-memory per-IP rate limit. On serverless this is per-instance,
+// not global, so it caps bursts against a hot instance rather than guaranteeing
+// a hard ceiling — enough to blunt casual flooding of the chat space without
+// pulling in an external store.
+const RATE_LIMIT_MAX = 5; // requests
+const RATE_LIMIT_WINDOW_MS = 60_000; // per minute, per IP
+const rateHits = new Map<string, number[]>();
+
+function isRateLimited(ip: string): boolean {
+  const now = Date.now();
+  const cutoff = now - RATE_LIMIT_WINDOW_MS;
+  const hits = (rateHits.get(ip) || []).filter((t) => t > cutoff);
+  hits.push(now);
+  rateHits.set(ip, hits);
+  return hits.length > RATE_LIMIT_MAX;
+}
+
+function clientIp(req: NextApiRequest): string {
+  const fwd = req.headers["x-forwarded-for"];
+  const raw = Array.isArray(fwd) ? fwd[0] : fwd;
+  return (raw?.split(",")[0].trim() || req.socket.remoteAddress || "unknown");
+}
+
+// Strip characters that carry meaning in Google Chat `text` messages so user
+// input can't inject formatting (*bold* / _italic_), fake clickable links
+// (<url|label>), or break the layout with newlines. Also caps length.
+function sanitize(value: string, max = 200): string {
+  return value
+    .replace(/[<>|*_`\r\n]/g, " ")
+    .replace(/\s+/g, " ")
+    .trim()
+    .slice(0, max);
+}
+
+// Track the "webhook not configured" warning so it's logged once per instance
+// instead of on every submit — the off state is expected until the env is set.
+let warnedMissingWebhook = false;
 
 export default async function handler(
   req: NextApiRequest,
@@ -31,9 +72,14 @@ export default async function handler(
     unknown
   >;
 
-  // Honeypot — silently accept (and drop) bot submissions.
+  // Honeypot — the form renders a hidden `company_website` input that humans
+  // never see; a filled value means a bot. Silently accept and drop.
   if (data.company_website) {
     return res.status(200).json({ ok: true, delivered: false });
+  }
+
+  if (isRateLimited(clientIp(req))) {
+    return res.status(429).json({ ok: false, error: "rate_limited" });
   }
 
   const name = String(data.fullName ?? "").trim();
@@ -45,30 +91,35 @@ export default async function handler(
   }
 
   const lead = {
-    name,
-    email,
-    company: String(data.companyName ?? "").trim(),
-    page: String(data.page ?? ""),
+    name: sanitize(name, 120),
+    email: sanitize(email, 254),
+    company: sanitize(String(data.companyName ?? ""), 160),
+    page: sanitize(String(data.page ?? ""), 500),
     submittedAt: new Date().toISOString(),
   };
 
   const webhook = process.env.GOOGLE_CHAT_WEBHOOK_URL;
   if (!webhook) {
     // No PII in logs. Set GOOGLE_CHAT_WEBHOOK_URL to deliver leads to the space.
-    console.error(
-      "[blog-lead] GOOGLE_CHAT_WEBHOOK_URL is not configured — lead accepted but NOT delivered. Set the env var to enable delivery.",
-    );
+    // Expected steady state until the env is set, so warn once, not every submit.
+    if (!warnedMissingWebhook) {
+      warnedMissingWebhook = true;
+      console.warn(
+        "[blog-lead] GOOGLE_CHAT_WEBHOOK_URL is not configured — leads accepted but NOT delivered. Set the env var to enable delivery.",
+      );
+    }
     return res.status(200).json({ ok: true, delivered: false });
   }
 
   try {
-    // Google Chat renders *bold* / _italic_ and <url|label> in `text` messages.
+    // page is rendered as plain (sanitized) text, not a <url|label> link, so a
+    // direct POST can't inject a misleading clickable URL.
     const text =
       `*📨 New Keploy blog subscriber*\n` +
       `*Name:* ${lead.name}\n` +
       `*Email:* ${lead.email}\n` +
       `*Company:* ${lead.company || "—"}\n` +
-      `*Page:* ${lead.page ? `<${lead.page}|${lead.page}>` : "—"}\n` +
+      `*Page:* ${lead.page || "—"}\n` +
       `*Submitted:* ${lead.submittedAt}`;
 
     const chatRes = await fetch(webhook, {
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -34,6 +34,13 @@ import type { NextApiRequest, NextApiResponse } from "next";
 //     ping ("new blog lead — see dashboard") and keep identity only in the two
 //     stores above.
 
+// Cap how long this function may run. The delivery fetch is aborted at 5s
+// (see the AbortController below); this ceiling has to sit ABOVE that abort, or
+// the platform kills the invocation before the abort can fire and the catch —
+// the branch that logs a stuck webhook — never runs. 10s leaves headroom for
+// the abort to trip, log, and respond, and is within every Vercel plan's limit.
+export const maxDuration = 10;
+
 const EMAIL_RE = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
 
 // `source` is pinned to a known set server-side — never echoed from the request
@@ -47,7 +54,11 @@ const DEFAULT_SOURCE = "blog-newsletter";
 // pulling in an external store.
 const RATE_LIMIT_MAX = 5; // requests
 const RATE_LIMIT_WINDOW_MS = 60_000; // per minute, per IP
-const RATE_MAP_MAX_KEYS = 10_000; // evict stale IPs past this so the map can't grow unbounded
+// Soft cap on tracked IPs. Only IPs idle past the window are evictable, so a
+// flood rotating more than this many IPs *within* a single window can still push
+// the map past it — acceptable for an in-memory best-effort limiter (the real
+// ceiling against distributed abuse is the reCAPTCHA-verification follow-up).
+const RATE_MAP_MAX_KEYS = 10_000;
 const rateHits = new Map<string, number[]>();
 
 function isRateLimited(ip: string): boolean {
@@ -63,6 +74,9 @@ function isRateLimited(ip: string): boolean {
   }
 
   const hits = (rateHits.get(ip) || []).filter((t) => t > cutoff);
+  // Record this hit even when it's the one that trips the limit. Deliberate: a
+  // sustained flooder keeps their own window sliding forward and stays limited,
+  // rather than earning a fresh allowance the instant they pause for a beat.
   hits.push(now);
   rateHits.set(ip, hits);
   return hits.length > RATE_LIMIT_MAX;
@@ -139,7 +153,13 @@ export default async function handler(
   }
 
   const name = String(data.fullName ?? "").trim();
-  const email = String(data.email ?? "").trim();
+  // Sanitize the email BEFORE validating, so the value we check is the value we
+  // deliver. EMAIL_RE accepts `*`, `|` and `` ` `` in a local part, but
+  // sanitize() turns each into a space — so validating the raw value could pass
+  // an address that sanitize then mangles (a*b@x.com -> "a b@x.com"), delivering
+  // a broken *Email:* line. Validating the sanitized form 400s those instead,
+  // and never harms real leads (`_`/`~` survive sanitize, see its note).
+  const email = sanitize(String(data.email ?? "").toLowerCase(), 254);
   // Re-validate server-side: a request could hit this endpoint directly and
   // bypass the client-side checks.
   if (!name || !EMAIL_RE.test(email)) {
@@ -149,9 +169,9 @@ export default async function handler(
   const rawSource = String(data.source ?? "").trim();
   const lead = {
     name: sanitize(name, 120),
-    // Lowercase server-side too (the client already does), so a direct hit
+    // Already sanitized + lowercased above (before validation), so a direct hit
     // can't create case-variant leads — same normalization the blog-mql path uses.
-    email: sanitize(email.toLowerCase(), 254),
+    email,
     company: sanitize(String(data.companyName ?? ""), 160),
     page: sanitize(String(data.page ?? ""), 500),
     source: ALLOWED_SOURCES.has(rawSource) ? rawSource : DEFAULT_SOURCE,
@@ -175,7 +195,10 @@ export default async function handler(
   }
 
   const controller = new AbortController();
-  const timer = setTimeout(() => controller.abort(), 15000);
+  // Abort a stuck webhook well under maxDuration (10s) so the catch runs and
+  // logs while the function is still alive; a Chat webhook answers in <1s, so
+  // 5s is generous headroom, not a tight bound.
+  const timer = setTimeout(() => controller.abort(), 5000);
   try {
     // page is rendered as plain (sanitized) text, not a <url|label> link, and
     // every field is sanitized, so a direct POST can't inject markup.
@@ -208,11 +231,20 @@ export default async function handler(
     }
     return res.status(200).json({ ok: true });
   } catch (err) {
+    // A network-level failure (DNS / refused connection / TLS / the 15s abort)
+    // is just as silent and permanent as a 4xx — leads quietly stop arriving —
+    // so surface it in production too, matching the !chatRes.ok branch above.
+    // The message is a fixed, PII-free string that never carries the webhook
+    // URL; the raw error (PII-free itself for undici, but belt-and-suspenders)
+    // is only attached off production.
+    const aborted = err instanceof Error && err.name === "AbortError";
+    const msg =
+      `[blog-lead] delivery to Google Chat failed (${aborted ? "timed out" : "network error"}) — ` +
+      "verify GOOGLE_CHAT_WEBHOOK_URL is a valid incoming-webhook URL. Lead was NOT delivered.";
     if (process.env.NODE_ENV !== "production") {
-      console.warn(
-        "[blog-lead] delivery to Google Chat failed — verify GOOGLE_CHAT_WEBHOOK_URL is a valid incoming-webhook URL. Lead was NOT delivered.",
-        err,
-      );
+      console.warn(msg, err);
+    } else {
+      console.warn(msg);
     }
     // Never fail the user — the newsletter subscription path is unaffected.
     return res.status(200).json({ ok: true });
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -90,13 +90,18 @@ function clientIp(req: NextApiRequest): string {
 // or forge whole labelled fields with newlines. Also caps length. Every
 // user-supplied field passes through this before it reaches the message.
 //
-// `_` and `~` are deliberately NOT stripped: both are legal in email local parts
-// (RFC atext, which EMAIL_RE accepts) and `_` is common in UTM-tagged page URLs
-// (utm_source, q3_launch), so stripping them corrupts real lead data. Chat italic
-// (_text_) and strikethrough (~text~) each need a matching pair — a lone one can't
-// format — and the link/field-forgery vectors are already dead from removing
-// < > | * and newlines, so the worst a stray pair can do is cosmetic styling,
-// never forge a field or a link.
+// We strip `< > | * ` \r \n` but deliberately KEEP `_` and `~`. The test is NOT
+// "is it atext" — `|`, `*` and `` ` `` are atext too, yet we strip them, because
+// each injects something concrete: `|` completes a clickable <url|label>, `*`
+// bolds, and `` ` `` opens a code span. The distinction that actually earns `_`
+// and `~` a pass is two-sided:
+//   • Risk: with `< > |` and newlines already gone, all `_`/`~` can still do is
+//     cosmetic matched-pair styling (_italic_, ~strike~; a lone one can't format)
+//     — never a forged field or a forged link.
+//   • Cost: `_`/`~` are legal in email local parts (RFC atext, which EMAIL_RE
+//     accepts) and `_` is common in UTM page URLs (utm_source, q3_launch), so
+//     unlike `* | ` ` ``, they appear in real lead data and stripping them would
+//     corrupt real leads.
 function sanitize(value: string, max = 200): string {
   return value
     .replace(/[<>|*`\r\n]/g, " ")
--- a/pages/api/blog-lead-notify.ts
+++ b/pages/api/blog-lead-notify.ts
@@ -12,20 +12,80 @@ import type { NextApiRequest, NextApiResponse } from "next";
 //     the webhook URL is never exposed to the client, git, or logs.
 //   • Create it in Google Chat: open the space → Apps & integrations → Webhooks
 //     → add one → copy the URL into the env var. Rotate by deleting/recreating.
+//   • This is a public, unauthenticated POST whose side effect is a message in
+//     the team space, so every user-supplied field MUST be sanitized (see
+//     sanitize()) and abuse is blunted with a honeypot + per-IP rate limit.
+//     Stronger gating (reCAPTCHA Enterprise verification, matching /blog-mql) is
+//     tracked as a follow-up.
 // If the env var is unset the submission still succeeds for the user, but the
 // lead is NOT delivered.
 
 const EMAIL_RE = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;
 
-// Field caps mirror blog-mql.ts. Google Chat rejects `text` over ~4096 chars, so
-// an oversized page URL or a long paste would silently fail delivery (chatRes.ok
-// false) even for a legit user. Truncate client-visible fields to stay safe.
-const CAPS = { name: 120, email: 254, company: 160, page: 500 };
-const cap = (value: string, max: number) => value.slice(0, max);
+// `source` is pinned to a known set server-side — never echoed from the request
+// body — so a direct POST can't forge a trustworthy-looking `*Source:*` line.
+const ALLOWED_SOURCES = new Set(["blog-newsletter"]);
+const DEFAULT_SOURCE = "blog-newsletter";
 
-// Abort a hung Google Chat request so a slow webhook can't hold a serverless
-// slot open until the platform hard-timeout. Mirrors blog-mql.ts (15s).
-const CHAT_TIMEOUT_MS = 15000;
+// Best-effort in-memory per-IP rate limit. On serverless this is per-instance,
+// not global, so it caps bursts against a hot instance rather than guaranteeing
+// a hard ceiling — enough to blunt casual flooding of the chat space without
+// pulling in an external store.
+const RATE_LIMIT_MAX = 5; // requests
+const RATE_LIMIT_WINDOW_MS = 60_000; // per minute, per IP
+const RATE_MAP_MAX_KEYS = 10_000; // evict stale IPs past this so the map can't grow unbounded
+const rateHits = new Map<string, number[]>();
+
+function isRateLimited(ip: string): boolean {
+  const now = Date.now();
+  const cutoff = now - RATE_LIMIT_WINDOW_MS;
+
+  // Opportunistically drop IPs whose most recent hit is outside the window, so
+  // the map doesn't accumulate an entry per distinct IP for the instance's life.
+  if (rateHits.size > RATE_MAP_MAX_KEYS) {
+    rateHits.forEach((times, key) => {
+      if (times.length === 0 || times[times.length - 1] <= cutoff) rateHits.delete(key);
+    });
+  }
+
+  const hits = (rateHits.get(ip) || []).filter((t) => t > cutoff);
+  hits.push(now);
+  rateHits.set(ip, hits);
+  return hits.length > RATE_LIMIT_MAX;
+}
+
+// Identify the client for rate limiting. On Vercel the trustworthy client IP is
+// `x-real-ip` (set by the platform). Do NOT use the leftmost x-forwarded-for
+// token — that end is client-supplied and lets an attacker rotate it to dodge
+// the limit; the real IP is the LAST hop the trusted proxy appends.
+function clientIp(req: NextApiRequest): string {
+  const realIp = req.headers["x-real-ip"];
+  if (typeof realIp === "string" && realIp.trim()) return realIp.trim();
+
+  const fwd = req.headers["x-forwarded-for"];
+  const raw = Array.isArray(fwd) ? fwd[0] : fwd;
+  if (raw) {
+    const parts = raw.split(",").map((s) => s.trim()).filter(Boolean);
+    if (parts.length) return parts[parts.length - 1];
+  }
+  return req.socket.remoteAddress || "unknown";
+}
+
+// Strip characters that carry meaning in Google Chat `text` messages so user
+// input can't inject formatting (*bold* / _italic_), fake clickable links
+// (<url|label>), or forge whole labelled fields with newlines. Also caps length.
+// Every user-supplied field passes through this before it reaches the message.
+function sanitize(value: string, max = 200): string {
+  return value
+    .replace(/[<>|*_`\r\n]/g, " ")
+    .replace(/\s+/g, " ")
+    .trim()
+    .slice(0, max);
+}
+
+// Track the "webhook not configured" warning so it's logged once per instance
+// instead of on every submit — the off state is expected until the env is set.
+let warnedMissingWebhook = false;
 
 export default async function handler(
   req: NextApiRequest,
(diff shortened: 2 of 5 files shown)


dhananjay6561 (commented):
reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) (typescript + security guides). nice clean change overall, the fail-open design and keeping the webhook as a server only secret (not NEXT_PUBLIC) is exactly right, and re-validating on the server instead of trusting the client is good.

the one thing i'd want sorted before merge is that `/api/blog-lead-notify` is a fully open unauthenticated POST that pushes messages straight into the team chat space, with no recaptcha or rate limit. the sibling blog-mql path verifies a recaptcha enterprise token and this one doesn't, even though the client already mints a token. left inline notes on that plus a couple of smaller things (the honeypot never actually fires, and the chat text isn't escaped). details inline.

severity key: 🔴 blocking, 🟡 important, 🟢 nit.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+    return res.status(405).json({ ok: false, error: "method_not_allowed" });
+  }
+
+  const data = (req.body && typeof req.body === "object" ? req.body : {}) as Record<
+    string,
+    unknown
+  >;
+
+  // Honeypot — silently accept (and drop) bot submissions.
+  if (data.company_website) {
```
dhananjay6561: 🟡 [important] this honeypot can never fire. the form in subscribe-newsletter.tsx only renders fullName / email / companyName, nothing ever sends `company_website`, and i grepped the repo, the string only exists here. so right now it's dead code that reads like bot protection but isn't. either add a hidden `company_website` input to the form (visually hidden, `tabindex=-1`, `autocomplete=off`) so real bots trip it, or drop the check so it's not misleading.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+    return res.status(200).json({ ok: true, delivered: false });
+  }
+
+  const name = String(data.fullName ?? "").trim();
+  const email = String(data.email ?? "").trim();
+  // Re-validate server-side: a request could hit this endpoint directly and
+  // bypass the client-side checks.
+  if (!name || !EMAIL_RE.test(email)) {
```
dhananjay6561: 🔴 [blocking] this is an open unauthenticated endpoint and the only gate is a trivial email regex, so anyone can POST valid looking name+email in a loop and flood the team's chat space (notification spam, and see the message injection note below). the blog-mql path right next door verifies a recaptcha enterprise token server side, this one has nothing. since submitHandler already mints a token via `getToken('submit_lead')`, simplest fix is to forward that token in the body and verify it here before posting to chat. a basic per-ip rate limit would be the lighter alternative. as-is the 'honeypot + validation keeps bot noise out' claim in the PR body doesn't really hold.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  try {
+    // Google Chat renders *bold* / _italic_ and <url|label> in `text` messages.
+    const text =
+      `*📨 New Keploy blog subscriber*\n` +
+      `*Name:* ${lead.name}\n` +
+      `*Email:* ${lead.email}\n` +
+      `*Company:* ${lead.company || "—"}\n` +
+      `*Page:* ${lead.page ? `<${lead.page}|${lead.page}>` : "—"}\n` +
```
dhananjay6561: 🟡 [important] user controlled fields get interpolated straight into google chat markup with no escaping. `page` is the worst one, it's rendered as `<url|label>` and on a direct POST it's fully attacker controlled, so someone can inject a misleading clickable link, or break the layout / other fields by slipping a `|`, `>` or newline in. name/company can also inject `*`/`_` formatting. suggest stripping `<>|*_` and newlines from each field, or just render `page` as plain text instead of a link.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+    email,
+    company: String(data.companyName ?? "").trim(),
+    page: String(data.page ?? ""),
+    submittedAt: new Date().toISOString(),
+  };
+
+  const webhook = process.env.GOOGLE_CHAT_WEBHOOK_URL;
+  if (!webhook) {
+    // No PII in logs. Set GOOGLE_CHAT_WEBHOOK_URL to deliver leads to the space.
+    console.error(
```
dhananjay6561: 🟢 [nit] until the env var is set this logs at error level on every single submit, which will look like a recurring incident in vercel logs / any alerting. since 'not configured yet' is an expected steady state (per the PR body, it ships off), `console.warn` feels more accurate, or log it once.

components/subscribe-newsletter.tsx
```diff
@@ -108,6 +110,21 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean })
       });
     });
 
+    // Google Chat notification — fire-and-forget, fail-open. Forwards the lead
+    // to a Chat space via the server-side /api/blog-lead-notify handler (which
+    // holds the webhook secret). Never gates the subscription.
+    fetch(`${router.basePath || ''}/api/blog-lead-notify`, {
```
dhananjay6561: 🟢 [nit] this fetch sits outside the `getToken(...).then(...)` above, so it always fires even for submissions recaptcha would've scored as bot. if you end up verifying the token in the endpoint (see my note there), move this call inside that same `.then` and forward `recaptchaToken` so the two paths stay consistent.

pages/api/blog-lead-notify.ts
dhananjay6561, line 152: good catch, it was dead. added a real hidden company_website input to the form (off screen, tabindex -1, autocomplete off, aria-hidden) and it gets sent in the notify body, so the check actually fires now. fixed in c6733f6

pages/api/blog-lead-notify.ts
dhananjay6561: fixed. all fields are sanitized now (strip <>|*_ backticks and newlines, collapse whitespace, cap length) and page is rendered as plain text instead of a <url|label> link, so no injected formatting or fake clickable urls. c6733f6

pages/api/blog-lead-notify.ts
dhananjay6561: agreed, switched to console.warn and it only logs once per instance now instead of on every submit. c6733f6

pages/api/blog-lead-notify.ts
dhananjay6561, line 175: valid concern. did a proportionate fix in this pr and left the heavy part as a follow up:

- added a best effort per ip rate limit (5/min) so the space can't be trivially flooded
- the escaping fix on the other thread kills the message injection half of this

on full recaptcha verification: this repo only has the public site key, verifying the token needs the project id + api key secrets which live in telemetry, so wiring that here means provisioning + duplicating those secrets. also worth noting the landing /trial/submit endpoint is open+unauth the same way, so this isn't a regression this pr introduces. i'd rather do proper recaptcha verification as a focused follow up than expand this pr's scope. tracking it separately.

components/subscribe-newsletter.tsx
dhananjay6561, line 137: leaving the fetch where it is for now since we're not verifying the token server side in this pr (see the blocking thread, recaptcha verify is a follow up). when we do add verification i'll move this inside the same .then and forward the token so both paths stay consistent.

pages/api/blog-lead-notify.ts
dhananjay6561, line 175: 🔴 [blocking] this is a public unauthenticated POST whose side effect is a message in the team's chat space, and the only gate is an email regex anyone can satisfy in a loop. so it's a ready made chat spam / phishing vector (see the page-link note below). worth noting the rest of the repo already gates writes: revalidate.ts checks a shared secret, and the sibling blog-mql lead is verified with a recaptcha enterprise token that the client already mints here in submitHandler.

a shared secret won't work for a public form, and full recaptcha verification needs GCP creds server side which may not live in this project, so a realistic layered fix: (1) reject when the Origin / Referer isn't keploy.io (subscribeMutation already sends an Origin header, cheap and kills naive cross-site abuse), and (2) a small rate limit per ip. if GCP creds are available, forwarding + verifying the existing token is the stronger option. right now the PR body's 'honeypot + validation keeps bot noise out' doesn't really hold, see next comment.

pages/api/blog-lead-notify.ts
dhananjay6561, line 152: 🟡 [important] this honeypot can't ever fire. the form only renders fullName / email / companyName and the client only posts those, and grepping the repo `company_website` exists nowhere except this line. so it reads like bot protection but is dead code. either add a real hidden `company_website` field to the form (visually hidden, tabindex=-1, autocomplete=off) so bots that autofill everything trip it, or drop the check so it isn't misleading.

pages/api/blog-lead-notify.ts
dhananjay6561: 🟡 [important] user fields go straight into the chat text with no escaping, and `page` is the sharp edge: it's rendered as `<url|label>` and on a direct POST it's fully attacker controlled, so someone can drop a misleading clickable link into the team's space, or break the layout / neighbouring fields with a `|`, `>` or newline. name/company can also sneak in `*`/`_` formatting. proxy-image.ts already shows the house style for this (validate the url, https + allowlist). simplest here: strip `<>|*_` and newlines from each field, and either validate `page` is a keploy.io url or just print it as plain text instead of a link.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+    // Google Chat renders *bold* / _italic_ and <url|label> in `text` messages.
+    const text =
+      `*📨 New Keploy blog subscriber*\n` +
+      `*Name:* ${lead.name}\n` +
+      `*Email:* ${lead.email}\n` +
+      `*Company:* ${lead.company || "—"}\n` +
+      `*Page:* ${lead.page ? `<${lead.page}|${lead.page}>` : "—"}\n` +
+      `*Submitted:* ${lead.submittedAt}`;
+
+    const chatRes = await fetch(webhook, {
```
dhananjay6561, line 224: 🟡 [important] this `await fetch(webhook)` has no timeout, so if google chat hangs the serverless invocation stays open until the platform hard-timeout, tying up a function slot per stuck request. blog-mql.ts already solves this with an AbortController aborting at 15s, worth mirroring that here so a slow webhook can't pile up.

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const name = String(data.fullName ?? "").trim();
+  const email = String(data.email ?? "").trim();
+  // Re-validate server-side: a request could hit this endpoint directly and
+  // bypass the client-side checks.
+  if (!name || !EMAIL_RE.test(email)) {
+    return res.status(400).json({ ok: false, error: "validation" });
+  }
+
+  const lead = {
```
dhananjay6561, line 180: 🟡 [important] no length caps on any field. blog-mql.ts deliberately truncates (page 500, name 120, company 160) because unbounded fields cause downstream rejects, and the same applies here: google chat rejects text over ~4096 chars, so an oversized page url or a long paste makes the whole POST silently fail delivery (chatRes.ok false) even for a legit user. truncating each field the same way keeps it consistent and delivery-safe.

dhananjay6561 (commented):
second pass, this time reading it against the patterns already in this repo (blog-mql.ts, proxy-image.ts, revalidate.ts) rather than in isolation.

the good stuff holds up: webhook stays a server only secret (not NEXT_PUBLIC), fail-open so a chat hiccup never breaks a subscribe, server side re-validation instead of trusting the client, and routing through router.basePath is correct since the app is served under /blog. all good.

the theme of my comments is that this endpoint doesn't follow the conventions the neighbouring code already established: blog-mql caps field lengths + uses an abort timeout + gates logging on NODE_ENV, proxy-image validates outbound urls against an allowlist, revalidate gates its POST. this one does none of those, and it's a public write that fans out to the team chat space. inline below, ordered by severity.

severity key: 🔴 blocking, 🟡 important, 🟢 nit.

pages/api/blog-lead-notify.ts
dhananjay6561: 🟢 [nit] two things: this is `console.error` in production on every submit while the var is unset, and unset is the shipped default per the PR body, so logs / alerting will show a recurring error that's actually the expected steady state. also blog-mql.ts gates all its logging behind `NODE_ENV !== 'production'`. suggest matching that convention, or at least dropping to warn and logging once rather than per request.

components/subscribe-newsletter.tsx
dhananjay6561, line 137: 🟢 [nit] this fetch sits outside the `getToken(...).then(...)` just above, so it always fires even for submits recaptcha would score as a bot. if the endpoint ends up verifying a token (see my blocking note there), move this call inside that same `.then` and forward `recaptchaToken` so both lead paths carry the same signal.

pages/api/blog-lead-notify.ts
dhananjay6561: done, added an abortcontroller aborting at 15s (finally clears the timer), same as blog-mql.ts. 4490b1c

pages/api/blog-lead-notify.ts
dhananjay6561: agreed, this one's a real delivery-safety bug. capped name 120 / email 254 / company 160 / page 500, matching blog-mql.ts, so an oversized field can't push the text past the ~4096 limit. 4490b1c

pages/api/blog-lead-notify.ts
dhananjay6561: fixed, gated both logs on NODE_ENV != production and dropped to warn, matching blog-mql.ts. 4490b1c

pages/api/blog-lead-notify.ts
dhananjay6561: this one's stale now, the form does have the hidden company_website field (subscribe-newsletter.tsx, off screen + tabindex -1 + autocomplete off) and it's posted in the body, so the check actually fires. added it in an earlier commit and kept it through the revert.

pages/api/blog-lead-notify.ts
dhananjay6561: the sharp edge you flagged (page as <url|label>) isn't in the current code, page is rendered as plain text, not a link. char-stripping the other fields i'm leaving out for now: landing's endpoint doesn't escape either and the ask was to mirror it, plus with page already plain text + the new length caps the blast radius into an internal-only space is small. can revisit as a follow-up if we want it.

components/subscribe-newsletter.tsx
dhananjay6561: leaving as is, we're not verifying a token server side in this pr so there's nothing to forward yet. if recaptcha verification lands later i'll move this inside the same .then.

pages/api/blog-lead-notify.ts
dhananjay6561: not adding auth gating here in this pr, on purpose. this endpoint is a straight mirror of landing's /trial/submit which is also a public unauth post to a chat space, so it's the established pattern for this exact use case rather than a new hole. the origin/referer allowlist is cheap but risks silently dropping legit leads from preview/vercel domains, and full recaptcha verification needs the gcp project-id + api-key which live in telemetry, not this repo. i'd rather do recaptcha verification (forward + verify the token the client already mints) as a focused follow-up than half-gate it here. i did tighten the delivery-safety + convention gaps from your other comments (timeout, length caps, logging) in 4490b1c. will update the pr body claim so it doesn't overstate the protection.

dhananjay6561 (commented):
third pass, focused on test coverage and cross-path consistency (the earlier passes covered security + the handler internals).

the big one: there's already an e2e spec for this exact form, tests/e2e/LeadCapture.spec.ts, and this PR adds a new user-facing network call without touching it. it even has a fail-open case that's the perfect template for the fail-open behaviour you're adding. details inline. also confirmed from that spec's own comment that the sidebar (so this form) renders twice per page, desktop + mobile fallback, which is fine for submit but good to keep in mind.

components/subscribe-newsletter.tsx
```diff
@@ -108,6 +110,21 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean })
       });
     });
 
+    // Google Chat notification — fire-and-forget, fail-open. Forwards the lead
```
dhananjay6561, line 134: 🟡 [important] this new POST has no test coverage, and there's an obvious home for it: tests/e2e/LeadCapture.spec.ts already drives this exact form, stubs `**/blog-mql` + the recaptcha loader, and even has a 'fail open when recaptcha is unavailable' case. right now that spec doesn't stub `**/blog-lead-notify`, so as written the new fetch hits the real endpoint during the e2e run (it returns 200 with the webhook unset, so the test won't fail, but the behaviour is completely unexercised). worth adding a `page.route('**/blog-lead-notify', ...)` capture and asserting it fires with `{ fullName, email, companyName, page }`, mirroring the existing blog-mql assertions, plus a case that a notify failure doesn't block the subscribe (the fail-open promise this PR is built around).

pages/api/blog-lead-notify.ts
```diff
@@ -0,0 +1,88 @@
...
+    unknown
+  >;
+
+  // Honeypot — silently accept (and drop) bot submissions.
+  if (data.company_website) {
+    return res.status(200).json({ ok: true, delivered: false });
+  }
+
+  const name = String(data.fullName ?? "").trim();
+  const email = String(data.email ?? "").trim();
```
dhananjay6561: 🟢 [nit] small consistency gap with the blog-mql path: there the client lowercases the email and blog-mql records `source: 'blog-newsletter'` / `assetType: 'newsletter'`, but here the server only trims (a direct hit keeps mixed case) and the chat ping carries none of that source context. consider `.toLowerCase()` on email here too, and having the client pass `source` so the chat message can say where the lead came from, then the two destinations tell the same story.

components/subscribe-newsletter.tsx
dhananjay6561: good call, real gap. added it to LeadCapture.spec.ts in 040f089: stubbed **/blog-lead-notify (so the run stops hitting the real endpoint), asserted the notify fires with { fullName, email, companyName, source, page } alongside the blog-mql lead, and added a new case that a 500 from notify still lets the subscription lead go out (the fail-open promise). one honest note: i haven't run the full playwright suite locally, it typechecks clean and mirrors the existing blog-mql patterns in the same spec.

pages/api/blog-lead-notify.ts
dhananjay6561: done in 040f089. lowercased the email server-side too (client already did, but a direct hit kept mixed case), and the client now passes source: 'blog-newsletter' which the chat message prints as a *Source:* line, so the notify + blog-mql paths tell the same story.

dhananjay6561:
lgtm 👍 the core is clean, server only secret, fail-open, server side re-validation, and it's lint + type clean. shipping the notify side channel this way is fine.

only thing i'd still genuinely flag before this goes fully live is the open `/api/blog-lead-notify` endpoint being spammable into the chat space (the 🔴 note inline), an origin check + small rate limit would cover it. the rest of the inline comments are your call, none of them block the feature working.

amaan-bhati:

Re-review at `040f089` (4 files, +206 / -0), following the inline round on `1d21b2a`.

**Verified locally:** `npx tsc --noEmit` clean, `npm run test:unit` **34/34 pass** (unchanged count, see I2).

## 🚦 Verdict: 🔄 REQUEST CHANGES

| Severity | Count |
|---|---|
| 🔴 blocking | 2 |
| 🟡 important | 3 |
| 🟢 nit | 4 |
| 💡 suggestion | 1 |
| 🎉 praise | 5 |

## ✅ Prior round, verified at head

| Finding | Status |
|---|---|
| Dead honeypot | ✅ Real hidden `company_website` input added and posted |
| No length caps | ✅ Capped 120 / 254 / 160 / 500, matching `blog-mql.ts` |
| No fetch timeout | ✅ `AbortController` at 15s, cleared in `finally` |
| `console.error` on every submit | ✅ Gated on `NODE_ENV !== "production"`, dropped to `warn` |
| Chat markup injection | ❌ **Reverted**, see B1 / B2 |
| Public endpoint spammable | ❌ **Reverted**, see B1 |

## 🔴 Blocking

**B1. The rate limit and the Chat sanitizer were both reverted in `c3e9524`.**

`c6733f6` and `dbbe9dc` landed the two fixes for the sharpest findings of the last round. `c3e9524` ("refactor(blog): match landing repo's notify endpoint faithfully") then removed 81 lines, including:

- `isRateLimited()` plus `clientIp()`, the per-IP 5/min limit with map eviction
- the `sanitizeForChat` helper that stripped `< > | * _` backticks and newlines

Both are absent at `040f089`. Confirmed by grep: no `rate`, `limit`, `sanitiz`, or `escape` anywhere in `pages/api/blog-lead-notify.ts`.

The `clientIp()` implementation that was deleted was the careful one. It read `x-real-ip` and explicitly took the **last** `x-forwarded-for` token, with a comment on why the leftmost token is client-supplied and spoofable. That is the correct call on Vercel and it is now gone.

The stated reason is parity with the landing repo, which does not escape either. Parity is not a security argument. It makes both repos wrong rather than making this one right, and it is a strange trade when the correct code already existed in this branch two commits ago. Restore `dbbe9dc`'s versions of both.

**B2. Message injection is still live, and the `page` fix does not close it.**

The response on that thread was that `page` is now rendered as plain text rather than `<url|label>`, so the blast radius is small. `page` is indeed plain text now, but the vector was never specific to `page`: Google Chat interprets markup in the whole `text` field, so **any** unescaped field carries it. Reproduced against the code at head:

```
POST /blog/api/blog-lead-notify
{"fullName":"Legit User\n*Email:* ceo@keploy.io\n*Company:* <https://evil.example/verify|Approve this lead>",
 "email":"attacker@mail.com","companyName":"Acme", ...}
```

renders in the space as:

```
*📨 New Keploy blog subscriber*
*Name:* Legit User
*Email:* ceo@keploy.io
*Company:* <https://evil.example/verify|Approve this lead>   ← clickable, attacker-chosen label
*Email:* attacker@mail.com
*Company:* Acme
*Source:* blog-newsletter
```

Two things fall out of that. Interior newlines survive `trim()`, so an attacker forges whole labelled fields, and the forged pair sorts **above** the real one, so a reader scanning the message sees the attacker's values first. And `fullName` / `companyName` still produce a clickable link with an arbitrary label.

With B1 also open there is no rate limit and no token check, so this is a repeatable unauthenticated phishing primitive aimed at an internal channel. The fix is the helper that was already written and deleted, or switch the message to `cardsV2` with structured fields so markup is never interpreted.

## 🟡 Important

**I1. No in-flight guard, so each click sends another Chat message.**
`components/subscribe-newsletter.tsx`

`disabled={isSubscribeDisabled()}` only checks that the three fields are non-empty, and `submitHandler` sets no submitting state. Repeated clicks fire repeated notify POSTs. The subscription path hides this because it upserts by email, but Chat does not: the space gets one ping per click. A `submitting` flag on the button covers it.

**I2. The handler has zero test coverage.**

`tests/e2e/LeadCapture.spec.ts` stubs the route with `page.route('**/blog-lead-notify', ...)`, so the 115-line handler never executes in any test. The honeypot branch, the email regex, the caps, the no-webhook path, and the fail-open catch are all unexercised, and the unit count is unchanged at 34. `tests/lib/revalidate.test.ts` already unit-tests an API route in this repo, so the pattern exists. Given B1/B2 concern logic inside this exact file, tests are what stop the next parity refactor from silently removing them again.

**I3. Lead PII now lands in a third destination with no retention story.**

Name, email and company go into Google Chat message history. Deleting a record from `subscriptions` or `blog_mql` does not touch it, so a GDPR erasure request now has a surface nobody owns and no process covers. Either decide the retention explicitly, or send a PII-free ping ("new blog lead, see dashboard") and keep the identity in the two stores that already hold it.

## 🟢 Nits

**N1.** `delivered: chatRes.ok` tells an unauthenticated caller whether the webhook is configured and whether their message landed. That is a useful oracle for someone probing the endpoint. Return a constant `{ ok: true }`.

**N2.** The honeypot is client-rendered, so a direct POST just omits `company_website`. Good against autofill bots, no help against targeted abuse. Fine as-is, it just should not be counted as the abuse control, which is the gap B1 leaves.

**N3.** `source` is echoed from the request body (80 chars, attacker-controlled on a direct hit) and rendered as a labelled field, so a forged `*Source:*` reads as trustworthy provenance. Pin it server-side to a known set instead.

**N4.** The comment above the message builder documents the exact hazard ("Google Chat renders `*bold*` / `_italic_` and `<url|label>` in `text` messages") but only uses it for the template's own formatting. Worth stating that user fields must be stripped, so the next refactor does not re-remove the guard.

## 💡 Suggestion

**S1.** There is an architecture that deletes this whole class of problem. Telemetry's `/blog-mql` already receives every lead and already verifies the reCAPTCHA token server-side. Doing the Chat notify there, server to server, removes the public endpoint, the rate-limit question, the injection surface, and the "forward the token later" follow-up in one move. Worth weighing against landing-repo parity, which is the only thing the current shape buys.

## 🎉 Praise

- The length caps are the best-reasoned change in the PR: tied to Chat's ~4096 limit, with the observation that an oversized field silently fails delivery for a legitimate user.
- `AbortController` at 15s with `clearTimeout` in `finally`, mirroring `blog-mql.ts`, is exactly right.
- Fail-open is real, not just claimed. The notify path never gates the subscription, and there is an e2e test that forces a 500 from the notify route and asserts the lead still fires.
- Server-side re-validation and `toLowerCase()` normalization, with a comment explaining that a direct POST bypasses the client checks.
- The e2e test asserting the notify payload alongside the `blog-mql` payload is the right level to catch a client-side regression.

## Recommended path

Restore both deletions from `c3e9524` (**B1**), which also closes **B2**. Add the in-flight guard (**I1**) and unit tests for the handler (**I2**). Make a call on **I3**. Consider **S1** before this goes live, since it makes B1 and the reCAPTCHA follow-up moot.

dhananjay6561:
thanks for the thorough second pass, fair catch on the revert. all addressed in 8d42e77 (+ retention doc in 0d51dbe):

**blocking**
- **B1** restored both deletions from dbbe9dc. the per-ip rate limit (5/min, map eviction) is back, and clientIp reads x-real-ip / the last x-forwarded-for token with the spoofing comment intact. agree parity isnt a security argument, this branch already had the right code.
- **B2** closed. every user field now goes through sanitize() (strips `<>|*_` backticks + newlines, collapses whitespace, caps), not just page. so newline-forged fields, clickable links and bold/italic can't reach the text. added a test that posts your exact repro payload and asserts the message stays 7 lines with none of it surviving.

**important**
- **I1** added a submitting flag, button disabled while in-flight and submitHandler early-returns, so repeat clicks dont fan out chat pings. reset in handleSubscribe finally.
- **I2** added tests/lib/blog-lead-notify.test.ts, 10 tests mirroring revalidate.test.ts: sanitize/injection, rate limit, honeypot no-leak, no-webhook, caps, pinned source, fail-open. unit count 34 to 44, all green.
- **I3** decided to keep the PII but document it (0d51dbe): the handler now spells out that the chat space is a third pii store, retention is governed by the space's message-retention policy (must be bounded), erasure requests have to purge chat too, and who owns it. left a note to switch to a pii-free ping if that ever becomes a burden.

**nits**
- **N1** all accept paths return a constant `{ ok: true }` now, no delivered oracle.
- **N3** source is pinned to a server-side allowlist, forged source falls back to blog-newsletter (tested).
- **N4** comment now says every user field must be sanitized so the next refactor doesnt re-remove it.
- **N2** agreed, honeypot isnt the abuse control, the restored rate limit is.

**S1** agree the server-to-server notify from /blog-mql is the cleaner shape and would delete this public endpoint + the recaptcha follow-up entirely. out of scope for this pr but noting it as the intended follow-up.

tsc + lint clean, 44/44 unit pass.

amaan-bhati:

Re-review at `0d51dbe` (5 files, +471 / -2). Delta since Iteration 2: `8d42e77`, `0d51dbe`.

**Verified locally:** `npx tsc --noEmit` clean, `npm run test:unit` **44/44 pass** (was 34).

## 🚦 Verdict: 💬 COMMENT

Both blockers are genuinely closed. Two data-correctness bugs below are cheap fixes that should land **before `GOOGLE_CHAT_WEBHOOK_URL` is set in Vercel**, since the feature is dormant until then. Nothing here blocks the merge.

| Severity | Count |
|---|---|
| 🔴 blocking | 0 |
| 🟡 important | 2 |
| 🟢 nit | 4 |
| 💡 suggestion | 1 |
| 🎉 praise | 5 |

## ✅ Iteration 2 findings, verified at head

| # | Finding | Status |
|---|---|---|
| B1 | Rate limit + sanitizer reverted in `c3e9524` | ✅ Restored. `isRateLimited` (5/min, map eviction) and `clientIp` are back, including the `x-real-ip` / last-XFF comment on spoofing |
| B2 | Injection live via any field | ✅ Closed. Every user field goes through `sanitize()`. Confirmed my repro payload now yields a 7-line message with no forged fields and no clickable link |
| I1 | No in-flight guard | ✅ `submitting` flag, button disabled, early return, reset in `.finally`. Confirmed `handleSubscribe` is `async`, so `.finally` is safe |
| I2 | Zero handler coverage | ✅ 10 tests in `tests/lib/blog-lead-notify.test.ts`, 34 to 44. Distinct-IP fixtures correctly isolate the module-level rate map |
| I3 | PII retention | ✅ Documented: third PII store named, bounded retention required, erasure must purge Chat, owner assigned |
| N1 | `delivered` oracle | ✅ Constant `{ ok: true }` on every accept path. But see T2 |
| N3 | Echoed `source` | ✅ Pinned to `ALLOWED_SOURCES`, with a test |
| N4 | Comment did not warn | ✅ Now states every user field must be sanitized |

## 🟡 Important

**T1. `sanitize()` strips `_`, which corrupts emails and UTM-tagged page URLs.**
`pages/api/blog-lead-notify.ts`

`_` is in the strip set and gets replaced with a space, then whitespace collapses. Two real consequences, both reproduced against the code at head:

```
john_doe@keploy.io                     → john doe@keploy.io
a_b.c_d@sub.keploy.io                  → a b.c d@sub.keploy.io

.../technology/x?utm_source=twitter&utm_campaign=q3_launch
→ .../technology/x?utm source=twitter&utm campaign=q3 launch
```

Underscores are legal in the local part and `EMAIL_RE` accepts them, so the address in the Chat message is wrong and the team cannot reply to that lead. That defeats the point of the notification for those submitters. The `page` case is worse in aggregate: `page` is `window.location.href`, so every campaign-tagged visit ships a broken attribution URL, and campaign leads are the ones this notification most exists for.

The tests do not catch it because the fixtures are `jane@keploy.io` and `https://keploy.io/blog/x`.

A lone `_` cannot produce italics without a matching pair in the same field, and the link and field-forgery vectors are already dead from removing `< > | *` and newlines. Dropping `_` from the strip set closes this with no loss. If you want belt and braces, strip only paired `_..._`.

**T2. The Chat response status is now discarded, so a broken webhook fails completely silently.**
`pages/api/blog-lead-notify.ts`

```ts
await fetch(webhook, { ... });
return res.status(200).json({ ok: true });
```

`fetch` only rejects on network-level failure, not on an HTTP error status. So a revoked webhook, a deleted space, a wrong URL, or a Chat quota rejection returns 4xx/5xx and this code treats it as success: no log even in dev, no metric, and the client response is a constant by design. Iteration 2's `delivered: chatRes.ok` at least surfaced it somewhere.

The failure mode is "leads quietly stop arriving and nobody notices", which is the worst shape for a notification channel. Keep the constant client response and check the status server-side:

```ts
const chatRes = await fetch(webhook, { ... });
if (!chatRes.ok && process.env.NODE_ENV !== "production") {
  console.warn("[blog-lead] Chat rejected the message", chatRes.status);
}
```

Worth considering whether this one should log in production too, since it is the only signal that delivery has broken.

## 🟢 Nits

**T3.** The double-submit guard reads `submitting` from the render closure, so two clicks in the same tick both see `false`. `disabled` covers it once React re-renders, so this is narrow. A `useRef` latch is the airtight version.

**T4.** `~strikethrough~` is still interpretable by Chat (tilde is not in the strip set). Cosmetic only, no link or field forgery, but free to add for completeness.

**T5.** Bare URLs in `name` / `company` still autolink in Chat, so a phishing URL can land in the space without a deceptive label. Far weaker than what B2 closed. If you want it shut, defang non-`page` fields.

**T6.** The honeypot check runs before the rate limit, so a bot that always sends `company_website` is never rate-limited. It delivers nothing, but still burns a serverless invocation per request. Moving `isRateLimited` above the honeypot costs nothing.

## 💡 Suggestion

**S1 (carried).** The per-instance rate limit scales the wrong way: Vercel adds instances under load, so the effective ceiling rises exactly when an attacker pushes harder. It is the right stopgap and it is documented honestly, so treat B1 as blunted rather than closed. The agreed follow-up (notify from telemetry's `/blog-mql`, which already verifies the reCAPTCHA token server-side) is still the actual fix and would delete this endpoint along with the whole problem class.

## 🎉 Praise

- Restoring `dbbe9dc` wholesale rather than rewriting it, comment included. Keeping the reasoning about why the leftmost `x-forwarded-for` token is spoofable is the part most people would have dropped.
- The injection test posts the exact repro payload from the review and asserts the message stays 7 lines, which pins the newline-forgery vector specifically rather than just checking that some escaping happens.
- The retention block does the thing most PRs skip: names the third store, states that deleting from the other two is not enough, requires bounded retention, and assigns an owner.
- Distinct-IP fixtures with an opt-in fixed IP for the rate-limit test is the right way to handle a module-level map in a shared test process.
- `warnedMissingWebhook` as a once-per-instance latch is a small thing done properly, and the comment explains why the off state is expected rather than alarming.

## Recommended path

Fix **T1** and **T2** before the env var goes into Vercel. **T3** to **T6** are optional. **S1** stays the real follow-up.

dhananjay6561:
thanks for the iteration 3 pass. fixed both the important ones in d9ea4a8:

t1: dropped `_` from the sanitize strip set. you were right, it was legal in email local parts and all over utm params so it was mangling the exact leads and attribution urls this thing exists for. chat italic needs a matching pair and the link plus field forgery vectors are already dead from stripping the other chars, so a stray underscore is cosmetic at worst. added a test with john_doe@ and a utm tagged page so the fixtures actually cover it now.

t2: capturing chatRes and checking status now. went with logging in prod too since it's the only signal a configured webhook has broken and the line is just the http status, no pii. client still gets the constant ok true so no delivery oracle and the user is never blocked. test asserts a non ok status gets logged while still returning 200.

tsc clean, unit tests 44 to 46.

leaving t3 to t6 for now since they're optional, and s1 stays the real follow up (notify from telemetry /blog-mql which already verifies recaptcha). happy to fold in t6 if you want it, it's basically free.

amaan-bhati:

At `d9ea4a8`. Delta since Iteration 3: one commit.
**Verified:** tsc clean, `test:unit` **46/46 pass** (was 44).

## 🚦 Verdict: 💬 COMMENT / 0 blocking, nothing left gating the webhook

| 🔴 0 | 🟡 0 | 🟢 4 | 💡 1 | 🎉 2 |
|---|---|---|---|---|

## ✅ Iteration 3 verified

**T1 fixed.** `_` dropped from the strip set. Confirmed against the code at head:

```
john_doe@keploy.io                              → john_doe@keploy.io
?utm_source=tw&utm_campaign=q3_launch           → ?utm_source=tw&utm_campaign=q3_launch
"Legit\n*Email:* ceo@… <https://evil/x|Approve>" → "Legit Email: ceo@… https://evil/x Approve"
```

Lead data intact, injection still dead. The comment explains why `_` is safe (italic needs a matching pair, and the link and field-forgery vectors are gone), so the next reader will not re-add it.

**T2 fixed.** `chatRes.ok` is checked and warns with the HTTP status. Logging in production is the right call here and the reasoning is stated: it is the only signal a *configured* webhook has broken, and the line carries no PII. Covered by a new test.

Both fixes were the exact two-line versions. That closes everything I raised that needed action before `GOOGLE_CHAT_WEBHOOK_URL` goes into Vercel.

## 🟢 Nits (all optional, unchanged from Iteration 3)

**T3.** Double-submit guard reads `submitting` from the render closure; `useRef` is the airtight version.
**T4.** `~strikethrough~` still interpretable by Chat. Cosmetic only.
**T5.** Bare URLs in `name`/`company` still autolink, without a deceptive label.
**T6.** Honeypot runs before the rate limit, so a bot always sending `company_website` is never limited. Delivers nothing, still burns an invocation.

## 💡 Suggestion

**S1 (carried).** The per-instance rate limit still scales the wrong way, since Vercel adds instances under load. Fine as a stopgap. Notifying from telemetry's `/blog-mql`, which already verifies the reCAPTCHA token, remains the real fix and deletes this endpoint.

## 🎉 Praise

- Both fixes landed as the minimal change with the reasoning written next to it, including why the previously-stripped character is safe to keep. That is what stops a fix from being reverted three commits later, which is exactly what happened earlier in this PR.
- Choosing to log the non-OK status in production, and justifying it on the PII content of the line rather than on convention, is the right way to make that call.

dhananjay6561:
addressed iteration 4 in 4a59efd. took the two that actually improve things, skipped the ones that dont change the outcome:

t3 (double-submit): done, this one matters since the whole point is one chat ping per click. moved the guard onto a useRef and check/set it synchronously. reading the submitting state or the disabled button sees a stale false on a fast double-click because both handlers fire before react re-renders, so two POSTs could slip through. the state still drives the disabled/opacity ui, the ref decides who runs.

t4 (~strikethrough): kept it, didnt strip, but extended the comment. ~ is valid in an email local part (rfc atext, and EMAIL_RE accepts it), so stripping it corrupts real lead data the exact same way stripping _ would. its cosmetic only and cant forge a field or link once < > | * and newlines are gone, so its the same call as _.

t5 (bare url autolink): left as is. the deceptive-label vector (<url|label>) is already dead since < > | are stripped, so a bare url just autolinks to where it visibly points. not a phishing surface.

t6 (honeypot before rate limit): left as is. reordering doesnt save the invocation, the function has already been invoked and runs to return either the 200 or the 429 regardless, so it doesnt address the burned-invocation cost thats the stated harm.

s1 (per-instance rate limit): agreed, its a stopgap. already flagged in the code comment and the PR body that notifying from telemetry's /blog-mql (which already verifies the recaptcha token) is the real fix and deletes this endpoint. tracked as a follow-up.

tsc clean, test:unit 46/46, next lint clean.

amaan-bhati:
# Claude Review Skill: Iteration 5

At `4a59efd` (1 new commit). Reviewed from the diff. **`tsc` / `test:unit` not run this round** (no disk space on the review machine), so treat CI as the source of truth for those.

| 🔴 0 | 🟡 0 | 🟢 1 | 💡 1 | 🎉 1 |
|---|---|---|---|---|

## ✅ Iteration 4 verified

**T3 fixed, correctly.** `submittingRef` is checked and set synchronously before any await, the state is left to drive the disabled/opacity UI, and both are reset in the same `.finally`. The comment names the exact failure (two handlers run before React re-renders, so the state and the disabled attribute both still read `false`). That is the right split of responsibilities between ref and state.

**T4: you were right, I was wrong.** I suggested stripping `~` for completeness. `~` is RFC 5322 atext, so it is legal in an email local part, and stripping it would have corrupted real addresses in exactly the way T1 did. Declining it and documenting why is the better call.

T5 and T6 remain open as optional nits, unchanged.

## 🟢 Nit

**F1. The stated rationale does not match the strip set.** The comment now argues `_` and `~` are kept because they are "legal in email local parts … so stripping them corrupts real lead data". By that test `|`, `*` and backtick should also be kept, since all three are atext too, and the code does strip them. The real distinction is narrower and worth saying instead: keep the atext characters that cannot inject on their own, strip the ones that can (`|` builds `<url|label>`, `*` bolds, backtick opens code). Otherwise the next reader applies the rule as written and removes a guard.

## 💡 Suggestion

**S1 (carried).** The per-instance rate limit still scales the wrong way, since Vercel adds instances under load. Fine as a stopgap. Notifying from telemetry's `/blog-mql`, which already verifies the reCAPTCHA token, remains the real fix and deletes this endpoint.

## 🎉 Praise

- Pushing back on T4 with the RFC rather than just applying it. A review suggestion that would have reintroduced the bug the previous round fixed is exactly the kind that should be argued with, and the argument was the correct one.

dhananjay6561:
thanks for the iteration 5 pass @amaan-bhati. fixed f1 in df949a1.

you nailed the flaw in the old comment, the "atext so keep" line proved too much since |, * and backtick are atext too and we strip all three. rewrote it on the real split: | * and backtick each inject on their own (link, bold, code span) so they go, while _ and ~ can only do cosmetic matched pair styling once < > | and newlines are already gone, and they actually turn up in real lead data (email local parts, utm page urls like utm_source) so stripping them would corrupt leads.

s1 is fair, the per instance limit does scale the wrong way as vercel adds instances. agreed the real fix is notifying from telemetry /blog-mql where the recaptcha token is already verified, which deletes this endpoint entirely. keeping it as the stopgap for this pr and it stays flagged as a follow up in the header.

and thanks for the note on t4, glad pushing back with the rfc was the right move rather than reintroducing the exact bug the round before had just fixed.

amaan-bhati:
# Claude Review Skill: Iteration 6

**Head:** `df949a1` (1 new commit). The delta was a comment fix, so this pass went at the delivery path's failure modes instead of re-reading the diff.

| 🔴 | 🟡 | 🟢 | 💡 |
|---|---|---|---|
| 0 | 2 | 2 | 1 |

## ✅ Iteration 5 verified

**F1 fixed, and the rewrite is better than what I asked for.** The comment now leads with the real test rather than the atext one, and splits it into risk and cost per character. Stating why `|`, `*` and `` ` `` fail that test is what stops the next reader from "simplifying" the strip set.

S1 acknowledged and kept flagged in the header as a follow-up. Agreed on the disposition.

---

## 🟡 N1. The `catch` branch goes quiet in production, but the `!chatRes.ok` branch beside it does not

`pages/api/blog-lead-notify.ts:200-213`. The non-OK branch logs unconditionally, with an explicit rationale: *"Logged in production too (unlike the missing-webhook warning): it's the only signal that a configured webhook has broken."* That reasoning is right. The adjacent `catch` is gated on `NODE_ENV !== "production"`, so in production it swallows:

- DNS failure / host unreachable for `chat.googleapis.com`
- connection refused, TLS failure
- the 15s abort (see N2)

Those are as silent and as permanent as a 4xx. A revoked webhook returning 404 is visible; the space becoming unreachable is not, which is the same "leads quietly stop arriving and nobody notices" outcome the non-OK branch was written to prevent.

I checked the obvious objection, that logging `err` in production might leak the webhook URL, since the header promises it *"never reaches the client, git, or logs"* and the URL carries `key`/`token` in its query string. It does not leak. Node's undici gives:

```
name: TypeError | message: fetch failed
cause: getaddrinfo ENOTFOUND <host>
contains the URL or its query params? no
```

So a PII-free production line is safe here. Matching the non-OK branch's shape would do it.

## 🟡 N2. The 15s abort is at or above the platform's own ceiling, so it cannot reliably fire

`blog-lead-notify.ts:177-178`. The route sets no `maxDuration`, and `vercel.json` has only `redirects` and `headers` with no `functions` block, so it runs on Vercel's default function duration, which is below or equal to 15s depending on plan (10s on Hobby, 15s on Pro). A hung webhook is therefore killed by the platform at or before the abort, so the `catch` never runs and nothing is logged, even in dev where N1's gate would allow it.

The 15s value only does its job if the function is guaranteed to outlive it. Either drop the abort to ~5s, which is far more than a Chat webhook needs and leaves headroom for the response, or set `maxDuration` on the route explicitly so the two numbers are chosen together rather than one of them inherited.

## 🟢 Nits (2)

**N3. `EMAIL_RE` runs before `sanitize`, so the delivered `*Email:*` line can hold something that is not an email.** `blog-lead-notify.ts:143,152`. `EMAIL_RE` is `/^[^@\s]+@[^@\s]+\.[^@\s]+$/`, which accepts `*`, `|` and `` ` `` in a local part; `sanitize` then turns each into a space, so `a*b@x.com` validates and is delivered as `a b@x.com`. This is a separate point from whether those characters should be stripped, which F1 settled correctly. It's about order: validating the sanitized value instead would return a 400 rather than posting a broken address. Rare in practice, one line to make impossible.

**N4. A rate-limited request still records a hit.** `blog-lead-notify.ts:65-68`. `hits.push(now)` happens before the `> RATE_LIMIT_MAX` return, so a sustained flooder keeps extending their own window and growing their array (bounded by the window, so not a leak). Punishing the flooder is a reasonable design, but right now it reads as incidental. One clause saying it's deliberate would settle it, or return before the push if it isn't.

## 💡 N5. The map cap can be exceeded by the case it exists for

`blog-lead-notify.ts:58-63`. Eviction runs only when `size > RATE_MAP_MAX_KEYS` and deletes only keys whose last hit is already outside the window, so a caller rotating more than 10k IPs *inside* one window is exactly the case that cannot be pruned. Consistent with the documented best-effort framing, and S1 (notify from telemetry `/blog-mql`, deleting this endpoint) subsumes it, so noting rather than asking.

---

## ✅ Verified this round

| Check | Result |
|---|---|
| `test:unit` | 46/46 pass |
| `tsc --noEmit` | Clean |
| Rate limit boundary | 5 allowed, 6th is 429; matches the test and the `RATE_LIMIT_MAX = 5` comment |
| Chat 4096-char limit | Worst case is ~1.1 KB across the caps (120+254+160+500 + labels), so the cap test's premise holds |
| `finally { clearTimeout(timer) }` | Reached on every path including the early `!chatRes.ok` return; no leaked timer |
| Sanitizer claim, "`_`/`~` can only do cosmetic matched-pair styling" | Holds. Field forgery needs a newline and link forgery needs `<`, `>` or `|`; all are stripped, and `*` cannot be user-supplied so `*Email:*` can't be forged |
| Fetch error contents | Carries no URL or query params, so N1's fix does not risk the webhook secret |
| Branch state | 0 behind `origin/main`, 12 ahead |

Nothing here blocks. N1 is the one I would take, since it protects the same property the non-OK branch already argues for.

amaan-bhati:
# Iteration 6: local end-to-end verification

Ran the endpoint for real rather than reading it, since the last two rounds were static analysis. `df949a1`, isolated worktree, `next dev` on `:3111`.

`GOOGLE_CHAT_WEBHOOK_URL` pointed at a **local sink on `:4599`** that echoes each payload, so no traffic left the machine and no real Chat space was touched. Endpoint is `/blog/api/blog-lead-notify` (the `basePath: '/blog'`, which `subscribe-newsletter.tsx:137` gets right via `router.basePath`).

## Results

| # | Request | Response | Delivered? |
|---|---|---|---|
| 1 | `GET` | `405` + `Allow: POST` | no |
| 2 | `POST {}` | `400 {"ok":false,"error":"validation"}` | no |
| 3 | `POST` invalid email | `400 {"ok":false,"error":"validation"}` | no |
| 4 | `POST` + `company_website` (honeypot) | `200 {"ok":true}` | **no** |
| 5 | valid lead | `200 {"ok":true}` | yes |
| 6 | injection payload | `200 {"ok":true}` | yes, neutralized |
| 7 | `a*b@x.com` | `200 {"ok":true}` | yes, see N3 |
| 8 | 300-char name | `200 {"ok":true}` | yes, capped to exactly 120 |
| 9-16 | 8 identical POSTs | `200 x5`, then `429 x3` | 5 |

Sink recorded **10 hits total** across the session, which matches 1 warm-up + #5 + #6 + 2x #7/#8 + 5 from the flood. The `405`, both `400`s and the honeypot contributed zero, so the "silently accept and drop" claim holds at the wire level, not just in the unit test.

## What actually arrived in the Chat message

Valid lead:

```
*📨 New Keploy blog subscriber*
*Name:* Amaan Bhati
*Email:* amaan.bhati@keploy.io
*Company:* Keploy
*Source:* blog-newsletter
*Page:* /blog/technology/api-testing?utm_source=q3_launch
*Submitted:* 2026-08-20T16:20:52.190Z
```

Sent as `Amaan.Bhati@Keploy.IO`, delivered lowercased. `utm_source=q3_launch` kept its underscore, which is the F1 decision doing its job on a real request.

Injection attempt. Sent:

```json
{"fullName":"Evil*bold*\n*Email:* attacker@evil.com\n*Company:* Forged",
 "email":"a_b~c@x.com",
 "companyName":"<https://evil.com|CLICK ME>",
 "page":"/p?a=`code`&b=*x*",
 "source":"admin-override"}
```

Arrived:

```
*📨 New Keploy blog subscriber*
*Name:* Evil bold Email: attacker@evil.com Company: Forged
*Email:* a_b~c@x.com
*Company:* https://evil.com CLICK ME
*Source:* blog-newsletter
*Page:* /p?a= code &b= x
*Submitted:* 2026-08-20T16:20:52.224Z
```

Every vector is dead in a way that's now observable rather than argued:

- **Field forgery** collapsed onto one line. The forged `*Email:*` / `*Company:*` labels lost their `*` and their newlines, so they read as inert text inside the Name value and cannot be mistaken for real fields.
- **Fake link** `<https://evil.com|CLICK ME>` lost `<`, `>` and `|`, so it renders as plain text, not a clickable label.
- **Code span** and **bold** both defused.
- **`source: "admin-override"` was pinned** back to `blog-newsletter`.
- **`a_b~c@x.com` survived intact**, both the `_` and the `~`. This is the exact case F1's comment argues for, and it is now demonstrated rather than reasoned about.

## N1 confirmed: the production gap is real

`next dev` forces `NODE_ENV=development`, so I drove the handler directly with a mocked req/res (the shape `tests/lib/blog-lead-notify.test.ts` already uses) and flipped only `NODE_ENV`:

```
================= NODE_ENV=development =================
  network failure (catch branch)     HTTP 200  logged: YES -> [blog-lead] delivery to Google Chat failed...
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...

================= NODE_ENV=production ==================
  network failure (catch branch)     HTTP 200  logged: *** NOTHING ***
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...
```

Same code, one env var apart. In production a Chat 500 is visible and a DNS failure, refused connection, TLS error or abort is completely silent, while both return `200` to the user either way. That is the asymmetry N1 describes, and the fail-open behaviour is confirmed correct on both paths.

## N3 confirmed

`a*b@x.com` passes `EMAIL_RE` (`*` is neither `@` nor whitespace), then `sanitize` turns the `*` into a space:

```
*Email:* a b@x.com
```

The `*Email:*` line is delivering something that is not an email address. Validating the sanitized value instead would have returned a `400`. Rare input, one line to make impossible, and unrelated to the F1 question of *which* characters to strip.

## Unchanged

Verdict stays 💬 COMMENT, 0 blocking. The abuse controls, sanitizer, source pinning, caps and fail-open all behave exactly as documented under real requests. N1 moves from reasoned to reproduced; N2 (the 15s abort sitting at or above the platform ceiling) is the one thing here I can't demonstrate locally, since it needs Vercel's function timeout to bite.

tests/e2e/LeadCapture.spec.ts
amaan-bhati, line 121: @dhananjay6561 I think that there is a failing test, I think it is realted to the recaptcha failure, this is although not really possible to test locally, since there wont be any recaptcha token generated locally, we can only test this on prod, but i want to understand why gate with recaptcha here? Did we have this earlier on the newsletter or not? Why include recaptcha here? Do try and ideate with claude on this, lmk what you think about this. If not relevant anymore then also let me know.

amaan-bhati (commented):
@dhananjay6561 Have added a comment for the recaptcha test failure, explained properly in the comment, kindly address. Rest of the things lgtm, this is the only thing i am concerned about, Rest all LGTM!

amaan-bhati (commented):
Verified things on local end-to-end

Ran the endpoint for real rather than reading it, since the last two rounds were static analysis. `df949a1`, isolated worktree, `next dev` on `:3111`.

`GOOGLE_CHAT_WEBHOOK_URL` pointed at a **local sink on `:4599`** that echoes each payload, so no traffic left the machine and no real Chat space was touched. Endpoint is `/blog/api/blog-lead-notify` (the `basePath: '/blog'`, which `subscribe-newsletter.tsx:137` gets right via `router.basePath`).

## Results

| # | Request | Response | Delivered? |
|---|---|---|---|
| 1 | `GET` | `405` + `Allow: POST` | no |
| 2 | `POST {}` | `400 {"ok":false,"error":"validation"}` | no |
| 3 | `POST` invalid email | `400 {"ok":false,"error":"validation"}` | no |
| 4 | `POST` + `company_website` (honeypot) | `200 {"ok":true}` | **no** |
| 5 | valid lead | `200 {"ok":true}` | yes |
| 6 | injection payload | `200 {"ok":true}` | yes, neutralized |
| 7 | `a*b@x.com` | `200 {"ok":true}` | yes, see N3 |
| 8 | 300-char name | `200 {"ok":true}` | yes, capped to exactly 120 |
| 9-16 | 8 identical POSTs | `200 x5`, then `429 x3` | 5 |

Sink recorded **10 hits total** across the session, which matches 1 warm-up + #5 + #6 + 2x #7/#8 + 5 from the flood. The `405`, both `400`s and the honeypot contributed zero, so the "silently accept and drop" claim holds at the wire level, not just in the unit test.

## What actually arrived in the Chat message

Valid lead:

```
*📨 New Keploy blog subscriber*
*Name:* Amaan Bhati
*Email:* amaan.bhati@keploy.io
*Company:* Keploy
*Source:* blog-newsletter
*Page:* /blog/technology/api-testing?utm_source=q3_launch
*Submitted:* 2026-08-20T16:20:52.190Z
```

Sent as `Amaan.Bhati@Keploy.IO`, delivered lowercased. `utm_source=q3_launch` kept its underscore, which is the F1 decision doing its job on a real request.

Injection attempt. Sent:

```json
{"fullName":"Evil*bold*\n*Email:* attacker@evil.com\n*Company:* Forged",
 "email":"a_b~c@x.com",
 "companyName":"<https://evil.com|CLICK ME>",
 "page":"/p?a=`code`&b=*x*",
 "source":"admin-override"}
```

Arrived:

```
*📨 New Keploy blog subscriber*
*Name:* Evil bold Email: attacker@evil.com Company: Forged
*Email:* a_b~c@x.com
*Company:* https://evil.com CLICK ME
*Source:* blog-newsletter
*Page:* /p?a= code &b= x
*Submitted:* 2026-08-20T16:20:52.224Z
```

Every vector is dead in a way that's now observable rather than argued:

- **Field forgery** collapsed onto one line. The forged `*Email:*` / `*Company:*` labels lost their `*` and their newlines, so they read as inert text inside the Name value and cannot be mistaken for real fields.
- **Fake link** `<https://evil.com|CLICK ME>` lost `<`, `>` and `|`, so it renders as plain text, not a clickable label.
- **Code span** and **bold** both defused.
- **`source: "admin-override"` was pinned** back to `blog-newsletter`.
- **`a_b~c@x.com` survived intact**, both the `_` and the `~`. This is the exact case F1's comment argues for, and it is now demonstrated rather than reasoned about.

## N1 confirmed: the production gap is real

`next dev` forces `NODE_ENV=development`, so I drove the handler directly with a mocked req/res (the shape `tests/lib/blog-lead-notify.test.ts` already uses) and flipped only `NODE_ENV`:

```
================= NODE_ENV=development =================
  network failure (catch branch)     HTTP 200  logged: YES -> [blog-lead] delivery to Google Chat failed...
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...

================= NODE_ENV=production ==================
  network failure (catch branch)     HTTP 200  logged: *** NOTHING ***
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...
```

Same code, one env var apart. In production a Chat 500 is visible and a DNS failure, refused connection, TLS error or abort is completely silent, while both return `200` to the user either way. That is the asymmetry N1 describes, and the fail-open behaviour is confirmed correct on both paths.

## N3 confirmed

`a*b@x.com` passes `EMAIL_RE` (`*` is neither `@` nor whitespace), then `sanitize` turns the `*` into a space:

```
*Email:* a b@x.com
```

The `*Email:*` line is delivering something that is not an email address. Validating the sanitized value instead would have returned a `400`. Rare input, one line to make impossible, and unrelated to the F1 question of *which* characters to strip.

dhananjay6561:
I am putting a hold on this PR currently
Putting PR #410 and #411 on priority to get them merged and clear out the backlog.

Will address this review once the above PRs get merged.

dhananjay6561:
@amaan-bhati dug into the recaptcha thing, few things:

first, there isn't a failing test on the current head. the e2e (chromium) check is green on df949a1, the exact commit you reviewed, all four LeadCapture tests pass headless in CI. if you saw a red run it was on an earlier commit or a flake, current state is clean.

on "can't test locally without a real token": that's actually not needed here. the spec stubs grecaptcha in beforeEach, it fulfills the loader with an empty script and injects execute to return 'e2e-stub-token', so the token is a fake and the whole thing runs fully offline. that's why it passes in CI with no prod and no real google call. so it is testable, it's just stubbed rather than live.

on why recaptcha is here at all: it's not from this pr. it came in with #403 (the /blog-mql lead capture), already on main. that's what gates the lead path, client mints an invisible token and telemetry verifies it server side. this pr doesn't touch any recaptcha code, and the chat notify path i added doesn't use recaptcha at all, the notify fetch fires outside the getToken().then() so it carries no token. i only extended the existing #403 spec to also cover the notify call.

so nothing to fix here re recaptcha, it's pre-existing and still relevant to the blog-mql lead, just unrelated to what this pr changes.

separately, pushed 253920b addressing the iteration-6 review: log silent notify failures in prod too (n1), maxDuration=10 + 5s abort so they're picked together (n2), validate email after sanitizing (n3), and documented the rate-limit hit counting + map cap (n4/n5). lmk if you want me to walk through any of it.

dhananjay6561:
@amaan-bhati verified the two fixes the same way you did, drove the handler with a mocked req/res and flipped only NODE_ENV, against the patched code in 253920b.

n1, the prod gap you found is closed:

```
================= NODE_ENV=development =================
  network failure (catch branch)     HTTP 200  logged: YES -> [blog-lead] delivery to Google Chat failed (network error)...
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...

================= NODE_ENV=production ==================
  network failure (catch branch)     HTTP 200  logged: YES -> [blog-lead] delivery to Google Chat failed (network error)...
  HTTP 500 (!chatRes.ok branch)      HTTP 200  logged: YES -> [blog-lead] Google Chat rejected the message (HTTP 500)...
```

the prod catch branch that logged *** NOTHING *** in your run now logs, symmetric with the !chatRes.ok branch. the raw error is only attached in dev, so prod stays PII-free, checked the log for the webhook key/token and the full url, none of it leaks.

n3, a*b@x.com now gets rejected instead of delivered as 'a b@x.com':

```
a*b@x.com  -> HTTP 400 {"ok":false,"error":"validation"}  delivered: no
```

both behaviours are also locked into tests/lib/blog-lead-notify.test.ts so they can't regress (48/48 pass). n2 is the one that still can't be shown locally since it needs vercel's function timeout to bite, but maxDuration=10 now sits explicitly above the 5s abort so they're chosen together.

dhananjay6561:
Screenshot 2026-08-21 at 12 55 15 PM


verified n1 and n3 locally the same way you did, mocked req/res + flipped only NODE_ENV so i could hit the prod path (next dev forces development).

amaan-bhati:
## Verified locally at `253920b`. Holding approval on one line

Drove the handler directly with a mocked req/res and flipped only `NODE_ENV`, same approach as your screenshot, and checked the `maxDuration` export against Next's own config extractor.

`npm run test:unit` 48/48, `npx tsc --noEmit` clean.

| Finding | Status |
|---|---|
| N1 catch branch silent in production | ✅ fixed, verified in both envs |
| N3 `EMAIL_RE` before `sanitize` | ✅ fixed, verified, plus a test |
| N4 rate-limited request records a hit | ✅ documented as deliberate |
| N5 map cap exceedable within a window | ✅ documented, follow-up named |
| N2 abort vs function ceiling | ❌ abort fixed, but the `maxDuration` export is inert |

N1 and N3 are properly done. The one thing below needs fixing before I approve.

---

## ⛔ The blocker: `export const maxDuration = 10` is silently ignored here

This is a Pages Router API route. In Next 14.0.1 the bare export is read for **app** router only; for **pages** it comes solely from `export const config = {…}`. Next's source is explicit: `AUTHORIZED_EXTRA_ROUTER_PROPS` is read from the bare export when `pageType === "app"`, and from `config` when `pageType === "pages"`.

Confirmed against the extractor on this exact file:

```
pageType="pages"  as written (bare export)       extraConfig = {}
pageType="pages"  with export const config = {}  extraConfig = {"maxDuration":10}
pageType="app"    as written (bare export)       extraConfig = {"maxDuration":10}
```

```diff
-export const maxDuration = 10;
+export const config = { maxDuration: 10 };
```

To be clear about the severity: no request fails today. The abort is 5s and every Vercel plan's default ceiling is 10s or more, so "the abort trips before the platform kills the invocation" happens to hold anyway. What's wrong is that the comment above it states a guarantee that isn't actually in force, and the whole point of N2 was to make that guarantee explicit. If the abort is ever raised trusting the declared 10s, the catch stops logging again, which is precisely the silent-failure mode N1 just fixed. That's why I'd rather it be real than incidental.

Verified after applying it: 48/48 tests still pass, `tsc` clean.

---

## Two nits, take or leave

**`name` still has the pre-N3 ordering.** The fix landed for email but not the field beside it, so a required field can arrive blank:

```
fullName "***"      HTTP 200  "*Name:* "
fullName "<>|`"     HTTP 200  "*Name:* "
```

`!name` passes on the raw value, then `sanitize` maps every character to a space and collapses it away. No injection risk, just the same ordering one field over.

```diff
-  const name = String(data.fullName ?? "").trim();
+  // Sanitize before validating, same reason as the email below: sanitize() maps
+  // < > | * ` and newlines to spaces, so a raw-passing value like "***" would
+  // collapse to "" and ship a blank *Name:* line despite the non-empty check.
+  const name = sanitize(String(data.fullName ?? ""), 120);
...
   const lead = {
-    name: sanitize(name, 120),
+    name,
```

With that applied:

```
fullName "***"          HTTP 400  (not delivered)
fullName "<>|`"         HTTP 400  (not delivered)
fullName " Jane "       HTTP 200  "*Name:* Jane"
fullName "Ann O'Reilly" HTTP 200  "*Name:* Ann O'Reilly"
300-char name           HTTP 200  capped at 120
```

The apostrophe case is deliberate: `'` isn't in the strip set, so real names are unaffected.

**The catch comment says "the 15s abort"** while the abort is now `5000` (lines 234 vs 201). Introduced by this commit.

---

## The evidence on what's already fixed

**N1, production observability.** All three failure modes now surface in production, the abort is distinguished from a network error, and the webhook secret never reaches the log:

```
=== NODE_ENV=development ===
  network   HTTP 200  logged=YES  secretLeak=no
            "[blog-lead] delivery to Google Chat failed (network error) — verify GOOGLE_CHAT_WEB…"
  abort     HTTP 200  logged=YES  secretLeak=no
            "[blog-lead] delivery to Google Chat failed (timed out) — verify GOOGLE_CHAT_WEBHOOK…"
  http500   HTTP 200  logged=YES  secretLeak=no

=== NODE_ENV=production ===
  network   HTTP 200  logged=YES  secretLeak=no
  abort     HTTP 200  logged=YES  secretLeak=no
  http500   HTTP 200  logged=YES  secretLeak=no
```

Production previously logged nothing for the network and abort cases. Fail-open is intact, every path still returns `200`. I asserted against the literal `key=` / `token=` values in the webhook URL rather than just the message shape, so the no-leak result is a real check.

**N3, sanitize before validate.**

```
email a*b@x.com     HTTP 400  (not delivered)
email a_b~c@x.com   HTTP 200  "*Email:* a_b~c@x.com"
email A.B@Keploy.IO HTTP 200  "*Email:* a.b@keploy.io"
```

The mangled-address case 400s instead of delivering `a b@x.com`, `_`/`~` still survive per the earlier decision, and server-side lowercasing still happens.

Happy to approve as soon as the `export const config` line lands.

amaan-bhati (commented):
## Re-checked at `4542438`. The blocker is unchanged

The two new commits are a `main` merge (which brought in #416), so nothing here addresses the previous round. Re-verified on this head rather than assuming:

**The blocker, still present.** `blog-lead-notify.ts:42` is still the bare export, and Next still ignores it for a Pages Router route:

```
pageType="pages"  extraConfig = {}
pageType="app"    extraConfig = {"maxDuration":10}
```

```diff
-export const maxDuration = 10;
+export const config = { maxDuration: 10 };
```

Line 198's comment still promises `maxDuration (10s)` is in force. It isn't.

**Both nits also unchanged.** `name` still validates raw and delivers sanitized (`:155` / `:171`):

```
fullName '***'     HTTP 200  "*Name:* "
fullName '<>|`'    HTTP 200  "*Name:* "
fullName 'Jane'    HTTP 200  "*Name:* Jane"
```

And `:234` still reads "the 15s abort" while the abort is `5000`.

**The merge itself is clean:** `npm run test:unit` **52/52** (up from 48, picking up #416's cases), `npx tsc --noEmit` clean. CI: build, E2E chromium, DCO pass; lighthouse still running.

Same as before — happy to approve once the `export const config` line lands. The other two are optional.

dhananjay6561:
hey you're looking at 4542438 which is the main merge commit, one behind. the actual fix is in 8f4ff8b sitting right on top of it and it's already on the remote head. can you take a pull and recheck at 8f4ff8b?

all three are in there: line 47 is now export const config = { maxDuration: 10 } with a note above explaining the pages vs app extractor thing so it doesn't creep back, name is sanitized before validate at line 165, and the stale 15s comment is now 5s at line 244. if you run the extractor again at that head you'll get extraConfig with maxDuration for pageType pages.

the mixup is just timing, the main merge got pushed to the branch while the last round was open and my fix went on after it, so it's the newest commit. fetch and you should see all three resolved.

amaan-bhati (approved):
## Approving. Full local verification at `8f4ff8b`

Verified the whole surface, not just the last delta. Isolated worktree, `next dev`, a **local sink on `:4601`** standing in for the Chat webhook so nothing left the machine and no real space was touched, plus headless Chromium for the UI path.

`npm run test:unit` **52/52** · `npx tsc --noEmit` clean · CI green on this head: build, lighthouse, E2E chromium, DCO.

### Rejection paths — nothing delivered

```
GET  (wrong method)              405  {"ok":false,"error":"method_not_allowed"}
POST {}                          400  {"ok":false,"error":"validation"}
POST invalid email               400  {"ok":false,"error":"validation"}
POST fullName "***"              400  {"ok":false,"error":"validation"}
POST email a*b@x.com             400  {"ok":false,"error":"validation"}
POST honeypot filled             200  {"ok":true}

sink hits from all six: 0
```

The two `400`s in the middle are the validate-after-sanitize ordering on **both** fields: each would previously have passed validation and delivered a mangled value.

### Delivery paths — what the sink actually received

```
*Name:* Amaan Bhati
*Email:* amaan.bhati@keploy.io                        <- sent Amaan.Bhati@Keploy.IO
*Page:* /blog/technology/x?utm_source=q3_launch       <- underscore preserved
```

Injection attempt, sent with forged `*Email:*` / `*Company:*` lines containing real newlines, a `<url|label>` link, a code span, bold, and `source: "admin-override"`:

```
*Name:* Evil bold Email: fake@evil.com Company: Forged   <- collapsed inline, no forged fields
*Email:* a_b~c@x.com                                     <- '_' and '~' both survive
*Company:* https://evil.com CLICK                        <- link defused
*Page:* /p?a= code &b= x                                 <- backtick + bold stripped
*Source:* blog-newsletter                                <- forged source pinned
```

Caps, measured on the delivered payload:

```
Name delivered length = 120   (sent 300)
Page delivered length = 500   (sent 600)
```

### Abuse controls

```
req 1..5 -> 200 {"ok":true}
req 6    -> 429 {"ok":false,"error":"rate_limited"}
req 7    -> 429 {"ok":false,"error":"rate_limited"}
```

### Observability, both envs

`next dev` forces development, so I drove the handler directly with a mocked req/res and flipped only `NODE_ENV`:

```
NODE_ENV=development
  network  HTTP 200  logged=YES  secretLeak=no  "…failed (network error) — verify GOOGLE…"
  abort    HTTP 200  logged=YES  secretLeak=no  "…failed (timed out) — verify GOOGLE_CHA…"
  http500  HTTP 200  logged=YES  secretLeak=no  "…rejected the message (HTTP 500) — …"

NODE_ENV=production
  network  HTTP 200  logged=YES  secretLeak=no
  abort    HTTP 200  logged=YES  secretLeak=no
  http500  HTTP 200  logged=YES  secretLeak=no
```

All three visible in production, abort distinguished from network error, fail-open intact on every path. I asserted against the literal `key=` / `token=` values in the webhook URL, not just the message shape, so `secretLeak=no` is a real check.

### Function ceiling

```
pageType="pages"  extraConfig = {"maxDuration":10}

:47   export const config = { maxDuration: 10 };
:211  const timer = setTimeout(() => controller.abort(), 5000);
```

Read back through Next's own static-info extractor, so the 10s ceiling is genuinely in force and the 5s abort sits under it — the catch can run and log before the platform reclaims the invocation.

### UI double-submit guard

```
same-tick x3 (the race the ref guards)   POSTs = 1
sequential x3, 400ms apart               POSTs = 2   gap = 888ms
```

The same-tick case is the one the ref exists for and it holds at exactly one POST. The sequential case is an in-flight guard behaving correctly — the second POST landed 888ms later, after the first had settled. Worth noting I could **not** verify whether the success state locks the form, because the newsletter subscription host isn't reachable from this environment, so `subscribed` never flips locally. In production that path is what would swap the UI.

---

Everything from the six rounds is closed and verified: the sanitizer contract and its `_`/`~` decision, source pinning, honeypot, per-IP rate limit, field caps, fail-open, production observability, validate-before-sanitize on both fields, and the function-ceiling guarantee now being real rather than incidental. Approving.

nehagup (commented):
## Review — newsletter form → Google Chat notify

**Recommendation: Approve with nits.** Well-engineered and hardened across the review iterations; the security-critical items all pass at head, and CI is green (E2E / build / Lighthouse / DCO).

### Security / abuse — passes
- **No committed secret** — the webhook is `process.env.GOOGLE_CHAT_WEBHOOK_URL` (never `NEXT_PUBLIC_*`); `.env.local.example` ships it commented-out as a placeholder. Error logs are PII- and secret-free.
- **Endpoint is defended** — honeypot, per-IP rate limit (5/min, with correct `x-real-ip` / last-`XFF` handling against spoofing), field caps, a `fetch` timeout, and **fail-open** so a Chat outage never blocks a subscription. User input is sanitized before it enters the Chat card.

### Should-fix
- **File the durable fix as a tracked issue, not a `// follow-up`.** The rate limiter is in-memory *per serverless instance*, so its effective ceiling rises as Vercel scales out under load. The real mitigation — send this notification server-to-server from the already-reCAPTCHA-verified `/blog-mql` path and delete this public endpoint — is only referenced in comments. Reasonable stopgap for an internal-channel notifier, but "follow-up" should be a real ticket.

### Process — worth reconciling before merge
The PR shows an `APPROVED` decision, but every formal review on the thread is `COMMENTED`, and the last one explicitly held approval pending the `maxDuration` fix. That fix landed in a later commit, but no re-approval was submitted afterward. Worth confirming the final commit is actually approved before merging on the strength of that state.

### Nits
- Rate-map eviction only prunes expired IPs above a 10k-entry threshold, so >10k IPs rotated *within* one 60s window could grow the map unbounded for the instance lifetime — add a hard cap / LRU if the `/blog-mql` refactor slips.
- Autolinking of bare URLs in the `name` / `company` fields inside the Chat message is **by design** — no deceptive-label vector (`<`, `>`, `|` are stripped) — noted for the record, no action.

### Scope
Clean — 517 of the 628 additions are the handler + 48 unit tests; the client change is small and entirely in service of the feature.

<sub>Related (not for this PR): this endpoint is better protected than landing's <code>/trial/submit</code>, which still lacks a honeypot / rate limit — worth backporting these defenses there.</sub>

PR #415, opened 2026-08-17
feat: track banner impressions (for per-banner CTR)

dhananjay6561:
## Summary

Adds **impression tracking** to the sidebar ad banners so we can measure real **CTR (clicks ÷ impressions)** per banner, not just raw clicks.

## What it does

- Fires a **viewable impression once** each time a banner scrolls **≥50% into view** (IntersectionObserver, one impression per page load — not on every scroll).
- Sends to both, carrying `banner_id` (same pattern as the existing `banner_click`):
  - **GA4:** `gtag("event", "banner_impression", { banner_id })`
  - **Clarity:** `clarity("set", "banner_shown", banner_id)`
- Works for the fallback card too, and no-ops safely if `IntersectionObserver`/analytics aren't available.

## Why

Raw clicks alone can mislead — a banner shown more often will get more clicks even if it converts worse. With impressions we get true per-banner CTR:

```
CTR = banner_click / banner_impression   (per banner_id)
```

## GA4 note (no new setup)

`banner_id` is already registered as a custom dimension, so `banner_impression` shows up with the same breakdown. In an Exploration you can put both `banner_click` and `banner_impression` counts side by side per `banner_id` and derive CTR.

## Testing

- Verified locally (headless): impression fires once on view — `dataLayer` gets `["event","banner_impression",{banner_id}]` and Clarity gets `["set","banner_shown",banner_id]`.
- `tsc` + `next lint` clean.

### DIFF
--- a/components/BlogSidebar.tsx
+++ b/components/BlogSidebar.tsx
@@ -1,6 +1,5 @@
 import React, { useState } from "react";
 import Link from "next/link";
-import Image from "next/image";
 import { useRouter } from "next/router";
 import SubscribeNewsletter from "./subscribe-newsletter";
 import {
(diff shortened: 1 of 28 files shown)


dhananjay6561:
closing to recreate this cleanly from an up-to-date main

PR #416, opened 2026-08-17
feat: track banner impressions (for per-banner CTR)

dhananjay6561:
## Summary

Adds **impression tracking** to the sidebar ad banners so we can measure real **CTR (clicks ÷ impressions)** per banner, not just raw clicks. Clicks (`banner_click` / `banner_clicked`) are untouched and already shipped — this PR only adds the impression side.

## What it does

- Fires a **viewable impression once** each time a banner scrolls **≥50% into view** (IntersectionObserver, one impression per page load — not on every scroll).
- Sends to both, carrying `banner_id` (same pattern as the existing `banner_click`):
  - **GA4:** `gtag("event", "banner_impression", { banner_id })`
  - **Clarity:** `clarity("set", "banner_shown", banner_id)`
- Works for the fallback card too, and no-ops safely if `IntersectionObserver`/analytics aren't available.

No numeric "count" is sent — each impression is a single event carrying only `banner_id`, and GA4 does the counting by tallying events per banner.

## Why

Raw clicks alone can mislead — a banner shown more often gets more clicks even if it converts worse. With impressions we get true per-banner CTR:

```
CTR = banner_click / banner_impression   (per banner_id)
```

## New file: `lib/banner-rotation.ts`

The sidebar renders **twice per page** (`post-body.tsx` mounts a ≥1440px copy and a sub-1440px copy, one hidden via `display:none`). That was harmless when we only tracked clicks — only the visible copy gets clicked. But for impressions, two independently-mounted instances would each pick their own banner and fire their own impression, and resizing across the 1440px breakpoint mounts the hidden copy, lazy-loads its image, and fires a **second** impression for a **different** `banner_id` — inflating the denominator and understating CTR.

So the banner list, the per-page pick, and the impression latch now live in shared **module scope**: both instances agree on **one** banner and fire **exactly one** impression per page view. This is safe because the sidebar is imported with `dynamic(..., { ssr: false })` (client-only, one module instance per tab, never shared across requests). It also makes the pick unit-testable by injecting `rand` — see `tests/lib/banner-rotation.test.ts`.

## Ready-poll timing

GA (`afterInteractive`) and Clarity (`lazyOnload`) may not exist yet when we want to send, so `whenReady` polls until the global appears, then gives up. The click vs impression timeouts are deliberately asymmetric: clicks follow a human action so the scripts have long loaded (~5s is plenty), while impressions can fire within the first seconds on a wide desktop and race Clarity's `lazyOnload` on a slow connection — so impressions poll much longer (~20s) to avoid a **one-sided drop that would inflate CTR**. The long poller is aborted on unmount, so an impression never lands on the next page after an SPA nav.

## GA4 note (no new setup)

`banner_id` is already a registered custom dimension, so `banner_impression` shows up with the same breakdown alongside `banner_click`.

## Testing (verified in a real browser, headless)

- **Network `/collect`:** exactly one `banner_impression` reaches GA with the correct `banner_id`; no duplicate after scrolling away and back.
- **dataLayer hook:** count is `0` before the banner is seen, `1` once ≥50% in view, and stays `1` after scrolling away+back 3×.
- Clarity `banner_shown` fires too. `tsc` + `next lint` clean.

### DIFF
--- a/components/BlogSidebar.tsx
+++ b/components/BlogSidebar.tsx
@@ -124,20 +124,26 @@ declare global {
 }
 
 // Clarity (lazyOnload) and GA (afterInteractive) may be undefined on an early
-// click, so run each call once its global exists — poll briefly, then give up.
-function whenReady(get: () => unknown, use: () => void) {
-  if (typeof window === "undefined") return;
+// call, so run each once its global exists — poll briefly, then give up. Pass a
+// signal to drop the poller on unmount (impressions fire before the user acts,
+// so their poller can outlive the component; aborting avoids firing after an
+// SPA nav, which would attribute the event to the next page).
+function whenReady(get: () => unknown, use: () => void, signal?: AbortSignal) {
+  if (typeof window === "undefined" || signal?.aborted) return;
   if (typeof get() === "function") { use(); return; }
   let tries = 0;
   const timer = setInterval(() => {
     tries += 1;
-    if (typeof get() === "function") {
+    if (signal?.aborted) {
+      clearInterval(timer);
+    } else if (typeof get() === "function") {
       clearInterval(timer);
       use();
     } else if (tries >= 20) {
       clearInterval(timer); // ~5s elapsed; script never loaded (blocked?)
     }
   }, 250);
+  signal?.addEventListener("abort", () => clearInterval(timer));
 }
 
 function trackBannerClick(bannerId: string) {
@@ -145,9 +151,9 @@ function trackBannerClick(bannerId: string) {
   whenReady(() => window.clarity, () => window.clarity!("set", "banner_clicked", bannerId));
 }
 
-function trackBannerImpression(bannerId: string) {
-  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }));
-  whenReady(() => window.clarity, () => window.clarity!("set", "banner_shown", bannerId));
+function trackBannerImpression(bannerId: string, signal?: AbortSignal) {
+  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }), signal);
+  whenReady(() => window.clarity, () => window.clarity!("set", "banner_shown", bannerId), signal);
 }
 
 /* ── Ad / CTA Banner ── */
@@ -173,15 +179,18 @@ function SidebarAdBanner() {
     if (!banner || impressionFired.current || !(loaded || errored)) return;
     const el = containerRef.current;
     if (!el || typeof IntersectionObserver === "undefined") return;
+    // Abort any pending analytics-ready poller if we unmount (e.g. SPA nav)
+    // before gtag/clarity load, so the event never fires against the next page.
+    const ac = new AbortController();
     const io = new IntersectionObserver((entries) => {
       if (entries.some((e) => e.isIntersecting) && !impressionFired.current) {
         impressionFired.current = true;
-        trackBannerImpression(banner.id);
+        trackBannerImpression(banner.id, ac.signal);
         io.disconnect();
       }
     }, { threshold: 0.5 });
     io.observe(el);
-    return () => io.disconnect();
+    return () => { io.disconnect(); ac.abort(); };
   }, [banner, loaded, errored]);
 
   // Text fallback so the ad slot is never blank if the artwork fails to load
--- a/components/BlogSidebar.tsx
+++ b/components/BlogSidebar.tsx
@@ -3,6 +3,12 @@ import Link from "next/link";
 import { useRouter } from "next/router";
 import SubscribeNewsletter from "./subscribe-newsletter";
 import {
+  AD_BANNERS,
+  pickBannerForPage,
+  hasFiredImpression,
+  claimImpression,
+} from "../lib/banner-rotation";
+import {
   FaFacebook,
   FaLinkedin,
 } from "react-icons/fa";
@@ -101,15 +107,12 @@ function SidebarShare() {
   );
 }
 
-// 4 banner variants for A/B rotation. Each artwork is an edge-to-edge card with
-// an empty bottom band; the CTA is built in code and overlaid there (per-variant
-// colors). btnCenter = that band's vertical center (% from the top).
-const AD_BANNERS = [
-  { id: "banner_1", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/banner1.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "88.5%" },
-  { id: "banner_2", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/banner2.webp", btnBg: "#ffffff", btnText: "#ED5D0F", btnCenter: "88.5%" },
-  { id: "banner_3", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/banner3.webp", btnBg: "#ED5D0F", btnText: "#ffffff", btnCenter: "88.5%" },
-  { id: "banner_4", src: "https://keploy-devrel.s3.us-west-2.amazonaws.com/landing/banner4.webp", btnBg: "#16324F", btnText: "#ffffff", btnCenter: "88.5%" },
-];
+// AD_BANNERS + the per-page banner pick and one-impression-per-page latch live
+// in lib/banner-rotation.ts — shared so the two sidebar instances (desktop +
+// sub-1440 copy) agree on one banner and don't double-count on a resize across
+// the 1440px breakpoint. Each artwork is an edge-to-edge card with an empty
+// bottom band; the CTA is built in code and overlaid there (per-variant colors).
+// btnCenter = that band's vertical center (% from the top).
 
 const CTA_HREF = "https://app.keploy.io/signin";
 // Shared artwork shape (not a fixed size) — reserves space to avoid layout shift.
@@ -123,12 +126,28 @@ declare global {
   }
 }
 
+const READY_POLL_MS = 250;
+// Clicks: a human decision precedes them, so gtag (afterInteractive) / clarity
+// (lazyOnload) have long loaded. ~5s is plenty.
+const CLICK_READY_TRIES = 20;
+// Impressions: fire on scroll-into-view with no user action, so on a wide
+// desktop the banner can be in view within the first seconds — racing Clarity's
+// lazyOnload on a slow connection. Give up much later than clicks so impressions
+// aren't dropped while clicks aren't; a one-sided drop would inflate CTR. This
+// is safe to run long because the poller is aborted on unmount (see the effect).
+const IMPRESSION_READY_TRIES = 80; // ~20s
+
 // Clarity (lazyOnload) and GA (afterInteractive) may be undefined on an early
-// call, so run each once its global exists — poll briefly, then give up. Pass a
-// signal to drop the poller on unmount (impressions fire before the user acts,
-// so their poller can outlive the component; aborting avoids firing after an
-// SPA nav, which would attribute the event to the next page).
-function whenReady(get: () => unknown, use: () => void, signal?: AbortSignal) {
+// call, so run each once its global exists — poll, then give up after maxTries.
+// Pass a signal to drop the poller on unmount (impressions fire before the user
+// acts, so their poller can outlive the component; aborting avoids firing after
+// an SPA nav, which would attribute the event to the next page).
+function whenReady(
+  get: () => unknown,
+  use: () => void,
+  opts: { signal?: AbortSignal; maxTries?: number } = {},
+) {
+  const { signal, maxTries = CLICK_READY_TRIES } = opts;
   if (typeof window === "undefined" || signal?.aborted) return;
   if (typeof get() === "function") { use(); return; }
   let tries = 0;
@@ -139,11 +158,11 @@ function whenReady(get: () => unknown, use: () => void, signal?: AbortSignal) {
     } else if (typeof get() === "function") {
       clearInterval(timer);
       use();
-    } else if (tries >= 20) {
-      clearInterval(timer); // ~5s elapsed; script never loaded (blocked?)
+    } else if (tries >= maxTries) {
+      clearInterval(timer); // script never loaded (blocked?)
     }
-  }, 250);
-  signal?.addEventListener("abort", () => clearInterval(timer));
+  }, READY_POLL_MS);
+  signal?.addEventListener("abort", () => clearInterval(timer), { once: true });
 }
 
 function trackBannerClick(bannerId: string) {
@@ -152,39 +171,43 @@ function trackBannerClick(bannerId: string) {
 }
 
 function trackBannerImpression(bannerId: string, signal?: AbortSignal) {
-  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }), signal);
-  whenReady(() => window.clarity, () => window.clarity!("set", "banner_shown", bannerId), signal);
+  const opts = { signal, maxTries: IMPRESSION_READY_TRIES };
+  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }), opts);
+  whenReady(() => window.clarity, () => window.clarity!("set", "banner_shown", bannerId), opts);
 }
 
 /* ── Ad / CTA Banner ── */
 function SidebarAdBanner() {
+  const router = useRouter();
   const [banner, setBanner] = React.useState<(typeof AD_BANNERS)[number] | null>(null);
   const [loaded, setLoaded] = React.useState(false);
   const [errored, setErrored] = React.useState(false);
   const containerRef = React.useRef<HTMLDivElement>(null);
-  const impressionFired = React.useRef(false);
 
+  // Pick from shared state keyed by page: both sidebar instances get the same
+  // banner, and a SPA nav to another post re-picks + re-arms the impression.
   React.useEffect(() => {
-    setBanner(AD_BANNERS[Math.floor(Math.random() * AD_BANNERS.length)]);
-  }, []);
+    setBanner(pickBannerForPage(router.asPath));
+  }, [router.asPath]);
 
   // Count a viewable impression once the picked banner scrolls ≥50% into view,
-  // a single time per load, so clicks / impressions gives a real CTR. We gate on
-  // (loaded || errored) so it counts only when the artwork (or the fallback card)
-  // has actually painted — a slow/blocked banner showing just the skeleton
-  // shouldn't count toward the denominator.
+  // a single time per page view, so clicks / impressions gives a real CTR. We
+  // gate on (loaded || errored) so it counts only when the artwork (or the
+  // fallback card) has actually painted — a slow/blocked banner showing just the
+  // skeleton shouldn't count toward the denominator.
   // `loaded`/`errored` are deps so the observer re-attaches once the artwork
   // paints, and to the swapped-in fallback card (different DOM node, same ref).
   React.useEffect(() => {
-    if (!banner || impressionFired.current || !(loaded || errored)) return;
+    if (!banner || hasFiredImpression() || !(loaded || errored)) return;
     const el = containerRef.current;
     if (!el || typeof IntersectionObserver === "undefined") return;
     // Abort any pending analytics-ready poller if we unmount (e.g. SPA nav)
     // before gtag/clarity load, so the event never fires against the next page.
     const ac = new AbortController();
     const io = new IntersectionObserver((entries) => {
-      if (entries.some((e) => e.isIntersecting) && !impressionFired.current) {
-        impressionFired.current = true;
+      // claimImpression() latches shared across both instances and returns true
+      // exactly once per page view, so only the first-seen banner is counted.
+      if (entries.some((e) => e.isIntersecting) && claimImpression()) {
         trackBannerImpression(banner.id, ac.signal);
         io.disconnect();
       }
--- a/components/BlogSidebar.tsx
+++ b/components/BlogSidebar.tsx
@@ -145,21 +145,46 @@ function trackBannerClick(bannerId: string) {
   whenReady(() => window.clarity, () => window.clarity!("set", "banner_clicked", bannerId));
 }
 
+function trackBannerImpression(bannerId: string) {
+  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }));
+  whenReady(() => window.clarity, () => window.clarity!("set", "banner_shown", bannerId));
+}
+
 /* ── Ad / CTA Banner ── */
 function SidebarAdBanner() {
   const [banner, setBanner] = React.useState<(typeof AD_BANNERS)[number] | null>(null);
   const [loaded, setLoaded] = React.useState(false);
   const [errored, setErrored] = React.useState(false);
+  const containerRef = React.useRef<HTMLDivElement>(null);
+  const impressionFired = React.useRef(false);
 
   React.useEffect(() => {
     setBanner(AD_BANNERS[Math.floor(Math.random() * AD_BANNERS.length)]);
   }, []);
+
+  // Count a viewable impression once the picked banner scrolls ≥50% into view
+  // (fires a single time per load, so clicks / impressions gives a real CTR).
+  React.useEffect(() => {
+    if (!banner || impressionFired.current) return;
+    const el = containerRef.current;
+    if (!el || typeof IntersectionObserver === "undefined") return;
+    const io = new IntersectionObserver((entries) => {
+      if (entries.some((e) => e.isIntersecting) && !impressionFired.current) {
+        impressionFired.current = true;
+        trackBannerImpression(banner.id);
+        io.disconnect();
+      }
+    }, { threshold: 0.5 });
+    io.observe(el);
+    return () => io.disconnect();
+  }, [banner, errored]);
 
   // Text fallback so the ad slot is never blank if the artwork fails to load
   // (content blocker, S3 hiccup, etc.). Keeps a working CTA + click tracking.
   if (banner && errored) {
     return (
       <div
+        ref={containerRef}
         className="rounded-2xl p-5 flex flex-col justify-center"
         style={{ width: '100%', aspectRatio: AD_ASPECT, backgroundColor: '#FFF4EE' }}
       >
@@ -188,6 +213,7 @@ function SidebarAdBanner() {
   // slot, show a skeleton until the image decodes, then fade the artwork in.
   return (
     <div
+      ref={containerRef}
       className="rounded-2xl overflow-hidden"
       style={{
         position: 'relative',
--- a/components/BlogSidebar.tsx
+++ b/components/BlogSidebar.tsx
@@ -184,11 +184,18 @@ function SidebarAdBanner() {
   const [errored, setErrored] = React.useState(false);
   const containerRef = React.useRef<HTMLDivElement>(null);
 
+  // Key on the path WITHOUT the #fragment. A TOC click uses history.replaceState
+  // directly (TableContents/post-body), so asPath only picks up the hash on a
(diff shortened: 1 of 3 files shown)


components/BlogSidebar.tsx
```diff
@@ -145,21 +145,46 @@ function trackBannerClick(bannerId: string) {
...
+  // Count a viewable impression once the picked banner scrolls ≥50% into view
+  // (fires a single time per load, so clicks / impressions gives a real CTR).
+  React.useEffect(() => {
+    if (!banner || impressionFired.current) return;
+    const el = containerRef.current;
+    if (!el || typeof IntersectionObserver === "undefined") return;
+    const io = new IntersectionObserver((entries) => {
+      if (entries.some((e) => e.isIntersecting) && !impressionFired.current) {
+        impressionFired.current = true;
+        trackBannerImpression(banner.id);
```
dhananjay6561: 🟢 [nit / confirm intended] the impression fires on the container being ≥50% viewable, but the container reserves the slot (aspectRatio) and shows the skeleton before the artwork decodes, and the errored fallback card uses the same ref. so an impression counts when the *slot* is seen, not when the artwork actually paints, meaning a slow / blocked / errored banner still counts toward the denominator and slightly deflates CTR. that's a defensible definition of 'viewable', just want to confirm it's the one you want. if you'd rather count artwork-actually-shown, gating on `(loaded || errored)` before firing would tighten it.

components/BlogSidebar.tsx
```diff
@@ -145,21 +145,46 @@ function trackBannerClick(bannerId: string) {
...
+        io.disconnect();
+      }
+    }, { threshold: 0.5 });
+    io.observe(el);
+    return () => io.disconnect();
+  }, [banner, errored]);
```
dhananjay6561: 🟢 [nit] the `errored` dep is doing something subtle, it's here so the observer re-attaches to the swapped-in fallback card div (different DOM node, same ref) when the image fails. correct, but worth a one line comment since it reads like an unused dep at a glance and someone may 'clean it up' later.

dhananjay6561 (commented):
reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) (react guide). clean little change, and it mirrors the existing trackBannerClick pattern nicely so the two events line up for the CTR math. effect cleanup disconnects the observer, the impressionFired ref guards against re-firing, and it no-ops safely when IntersectionObserver / analytics aren't there. lint clean.

one thing i specifically checked since it looked risky: BlogSidebar renders twice in post-body.tsx (desktop ≥1440px + a sub-1440 copy), so two SidebarAdBanner instances mount and each sets up its own observer. that's fine here because the hidden one is `display:none` (tailwind `hidden`), and IntersectionObserver never reports a display:none element as intersecting, so only the visible instance fires. no double counting. 👍

two small notes inline, both non-blocking.

severity key: 🔴 blocking, 🟡 important, 🟢 nit.

components/BlogSidebar.tsx
dhananjay6561: good call, tightened it in 2edcbf8. now gated on (loaded || errored) so it only counts when the artwork actually paints (or the fallback card shows), not the bare skeleton slot. a slow/blocked banner no longer inflates the denominator. re-tested: still fires exactly once, right after the image paints, no duplicate on scroll away+back.

components/BlogSidebar.tsx
dhananjay6561: done in 2edcbf8, added a comment explaining the loaded/errored deps (observer re-attaches once the artwork paints and to the swapped-in fallback node) so nobody strips them as unused later.

components/BlogSidebar.tsx
```diff
@@ -145,21 +145,46 @@ function trackBannerClick(bannerId: string) {
   whenReady(() => window.clarity, () => window.clarity!("set", "banner_clicked", bannerId));
 }
 
+function trackBannerImpression(bannerId: string) {
+  whenReady(() => window.gtag, () => window.gtag!("event", "banner_impression", { banner_id: bannerId }));
```
dhananjay6561: 🟢 [nit, pre-existing but amplified here] whenReady spins a setInterval that isn't tied to react's lifecycle, so if the analytics global isn't ready yet it polls for up to ~5s regardless of whether the component unmounts. clicks rarely hit that path (analytics is usually loaded by the time someone clicks), but impressions auto-fire on mount / scroll into view with no user action, and on a wide desktop the sidebar banner can be in view immediately at load, before gtag (afterInteractive) / clarity (lazyOnload) are ready. so this new caller is a lot more likely to leave a poller running, and if the reader SPA-navigates within those 5s it can fire the event after unmount (banner_id is still correct via the closure, but the GA page context at send time is the next page). not blocking, just the kind of thing that shows up as slightly-off attribution once impression volume is high. a shared 'fire once ready or drop on unmount' helper would fix both callers, out of scope for this PR.

dhananjay6561 (commented):
second pass, this time on lifecycle / runtime edges rather than the happy path. mostly came back clean, noting what i checked so it's on record:

- SPA nav: BlogSidebar is `dynamic({ ssr:false })`, so it remounts per post and impressionFired resets, one impression per post view, correct.
- react strictmode double-invoke: dev only, and the impressionFired ref + io.disconnect guard against a double fire anyway.
- IO occlusion: an impression can fire while the banner is covered by a sticky/overlay (IntersectionObserver v1 doesn't track actual visibility), but that's the same limitation the existing click tracking lives with, not a regression.

one real (minor) thing inline about the whenReady poller. nothing blocking.

severity key: 🔴 blocking, 🟡 important, 🟢 nit.

components/BlogSidebar.tsx
dhananjay6561: nice catch, fixed the amplified case in 5c06c07. whenReady now takes an optional AbortSignal, and the impression effect passes one and aborts it in cleanup. so if the reader spa-navigates within the ~5s poll window before gtag/clarity load, the poller gets dropped instead of firing against the next page. left clicks as-is since those are fire-and-forget from onClick with the component still mounted. agree a shared helper for both callers would be cleaner but out of scope here. re-tested: still fires exactly once.

dhananjay6561:
lgtm 👍 clean change, mirrors the existing banner_click pattern so the CTR math lines up, observer cleanup + impressionFired guard are correct, and it's lint + type clean.

checked the bit that looked risky (BlogSidebar renders twice, desktop + sub-1440) and it doesn't double count because the hidden copy is `display:none` and IntersectionObserver never reports that as intersecting. the inline notes are all 🟢 nits, none block, and the whenReady one isn't even introduced here. ship it.

amaan-bhati:

Re-review at `5c06c07` (1 file, +44 / -4). Delta since the inline round: `2edcbf8`, `5c06c07`.

**Verified locally:** `npx tsc --noEmit` clean, `npm run test:unit` **34/34 pass**.

## 🚦 Verdict: 💬 COMMENT

The code is correct for a fixed viewport. Everything below is about whether the resulting CTR number can be trusted, which is the only reason this PR exists.

| Severity | Count |
|---|---|
| 🔴 blocking | 0 |
| 🟡 important | 2 |
| 🟢 nit | 3 |
| 💡 suggestion | 1 |
| 🎉 praise | 3 |

## ✅ Prior round, verified at head

| Finding | Status |
|---|---|
| Impression counted on the bare skeleton slot | ✅ Gated on `(loaded \|\| errored)`, so it counts only once artwork paints or the fallback card shows |
| `errored` dep read as unused | ✅ Comment added explaining the observer re-attaches to the swapped-in fallback node |
| `whenReady` poller outliving unmount | ✅ `AbortSignal` threaded through, aborted in the effect cleanup |

I also checked the risk called out in the lgtm: `BlogSidebar` does mount twice (`post-body.tsx:622` and `:628`), but both wrappers hide with Tailwind `hidden` / `min-[1440px]:hidden`, which is `display: none`. IntersectionObserver never reports a `display: none` element as intersecting, so at any fixed width exactly one instance fires. That part holds. The problem is what happens when the width is not fixed, below.

## 🟡 Important

**F1. Crossing the 1440px breakpoint double-counts one page view as two impressions, for two different banners.**
`components/BlogSidebar.tsx`, `components/post-body.tsx:622,628`

Both `BlogSidebar` instances mount, and each `SidebarAdBanner` picks its own banner independently:

```ts
setBanner(AD_BANNERS[Math.floor(Math.random() * AD_BANNERS.length)]);
```

`impressionFired` is a per-instance ref, so the two instances share no latch. At a fixed width the hidden one never loads its lazy image, `loaded` stays false, and the impression effect early-returns. Resize across 1440px (or rotate a tablet, or open devtools) and the previously hidden instance enters layout, its `loading="lazy"` image fetches, `onLoad` sets `loaded`, the effect runs, the observer attaches, and it fires a second `banner_impression` with a **different** `banner_id`.

Net effect on the metric: one reader, one page view, two impressions split across two banners, and at most one click. The denominator inflates and the click lands on only one of the two ids, so per-banner CTR is understated for both. Narrow trigger, but it corrupts exactly the number this PR produces, and it is silent.

Cheapest fix is to hoist the banner pick and the fired latch out of the component (module scope keyed per page view, or a shared context), so both instances agree on one banner and one impression. Rendering a single `BlogSidebar` and moving it with CSS would remove the class entirely.

**F2. The 5s analytics-ready give-up systematically undercounts impressions relative to clicks, which inflates CTR.**
`components/BlogSidebar.tsx` `whenReady`

`whenReady` polls 20 times at 250ms and then gives up silently. That budget was written for clicks, and it is fine for clicks: a click needs a human decision, by which time `gtag` (afterInteractive) and `clarity` (lazyOnload) have long loaded.

Impressions are different. They fire as soon as the banner is scrolled into view, which on the wide-desktop layout can be within the first seconds of the page, competing with `lazyOnload` Clarity on a slow connection. When the poller gives up, `impressionFired.current` is already `true`, so there is no retry and the impression is lost for good.

Because clicks are effectively never dropped and impressions sometimes are, the loss is one-sided: **CTR reads higher than reality, and worse on slow connections**. It also lets the GA and Clarity denominators diverge, since the two `whenReady` calls can time out independently.

Two options: raise the impression budget well past Clarity's load (it is `lazyOnload`, so 5s is genuinely tight), or set `impressionFired` only once a call actually lands so a timed-out attempt can be retried on the next intersection.

## 🟢 Nits

**F3.** `impressionFired.current = true` is set inside the observer callback before either analytics call is attempted, which is what makes F2 unrecoverable. Worth a comment either way, since "latched but never sent" is a non-obvious state.

**F4.** `whenReady` registers `signal.addEventListener("abort", ...)` and never removes it, and `trackBannerImpression` calls `whenReady` twice with the same signal. Harmless here (the listeners die with the signal) but it will accumulate if `whenReady` is ever reused with a longer-lived signal.

**F5.** No test for the fire-once latch. The IO plumbing is awkward to unit test, but "one impression per page view" is the invariant the metric depends on and it is the thing F1 breaks. A Playwright assertion on a stubbed `window.gtag` call count would cover it, and #414 just set the precedent for testing this kind of logic in this repo.

## 💡 Suggestion

**F6.** The comment calls `AD_BANNERS` a "4 banner variants for A/B rotation", but the pick is `Math.random()` per mount with no per-user or per-session stickiness, so a reader sees different banners on consecutive articles. That is fine for aggregate per-banner CTR, which is what the PR title claims, but it does not support A/B conclusions about users, and the numbers will get read that way once they exist. Worth softening the comment to "random rotation" so nobody over-reads the dashboard later.

## 🎉 Praise

- Gating on `(loaded || errored)` rather than container visibility is the right call, and it is the difference between a CTR denominator that means something and one that counts blocked banners.
- The click path deliberately does not get an abort signal, and that asymmetry is correct: the CTA is `target="_blank"`, so the page stays mounted and the poller should survive. Easy to get wrong in the other direction.
- Threading `AbortSignal` through `whenReady` instead of adding a second ad-hoc cancellation path keeps one mechanism for both callers, and the comment explains the SPA misattribution it prevents.

## Recommended path

**F1** is the one worth fixing before the numbers get used for anything. **F2** decides whether the CTR is systematically optimistic, so worth a decision even if the answer is "acceptable". The rest are optional.

dhananjay6561:
thanks, both important ones are fixed in 8a37046.

f1: you were right, the two instances each had their own banner pick + impressionFired ref, so a resize across 1440px would mount the previously display:none copy, load its lazy image and fire a second impression for a different banner_id. moved the banner pick and the fire-once latch into lib/banner-rotation.ts at module scope, keyed per page view. now both instances get the same banner and claimImpression() returns true exactly once per page, and a spa nav re-picks + re-arms. went with the shared-state route over a single css-moved sidebar since the two placements are structurally different (sticky aside in the flex row vs centered block below), so unifying with css isn't clean.

f2: bumped the impression poll budget to ~20s (clicks stay at ~5s). clicks follow a human decision so gtag/clarity are always loaded by then, but impressions auto-fire on scroll-in and can race clarity's lazyOnload on a slow connection, so the old 5s give-up dropped impressions but never clicks, which is the one-sided loss that inflates ctr. safe to poll longer now since the poller already aborts on unmount, so it can't leak or misattribute.

f3/f4/f6: added a note that the latch fires on intersection, made the abort listener once-only, and softened the "A/B rotation" wording to "random rotation" so nobody reads the dashboard as a per-user A/B result.

f5: added unit tests for the one-banner/one-impression-per-page invariant (the thing f1 broke) in tests/lib/banner-rotation.test.ts. extracting the shared state made it unit-testable without the IO plumbing.

tsc + next lint clean, unit tests 34 to 38.

amaan-bhati:

At `8a37046`. Delta since Iteration 2: one commit, plus `lib/banner-rotation.ts`.
**Verified:** tsc clean, `test:unit` **38/38 pass** (was 34).

| 🔴 0 | 🟡 0 | 🟢 3 | 💡 0 | 🎉 2 |
|---|---|---|---|---|

## ✅ Iteration 2 verified

| # | Status |
|---|---|
| F1 resize double-counts two banners | ✅ Pick and latch moved to shared module state keyed per page view, so both instances agree on one banner and `claimImpression()` returns true once |
| F2 5s give-up inflates CTR | ✅ Split into `CLICK_READY_TRIES` (20) and `IMPRESSION_READY_TRIES` (80, ~20s), with the one-sided-drop reasoning stated |
| F3 latch set before send | ⚠️ Still latches first, but the 20s window makes it far narrower. See R3 |
| F4 abort listener never removed | ✅ `{ once: true }` |
| F5 no test for fire-once | ✅ `tests/lib/banner-rotation.test.ts`, with an injectable `rand` |
| F6 "A/B rotation" overclaim | ✅ Comment now says random rotation, not an A/B conclusion about users |

Traced the resize path: at <1440 the mobile copy fires and claims; on resize the desktop copy's effect hits `hasFiredImpression()` and returns early. One impression per page view, and a SPA nav re-keys and re-arms correctly.

## 🟢 Nits

**R1. The page key includes the hash.** `pickBannerForPage(router.asPath)` and `asPath` carries the fragment. `TableContents.tsx:109` and `post-body.tsx:274` both call `window.history.replaceState(null, null, '#id')` directly, which bypasses the Next router, so a TOC click does not re-key today. But a later back or forward fires `popstate`, Next resyncs `asPath` with the hash, the key changes, and the latch re-arms, so a second impression can fire for the same page view. Nothing about the banner depends on the fragment, so `router.asPath.split("#")[0]` removes the whole question.

**R2. The module state is only safe because the sidebar is `ssr: false`.** `pickBannerForPage` mutates module-level `let`s. That is fine while `BlogSidebar` is `dynamic(..., { ssr: false })`, but if it is ever server-rendered the pick and the latch become shared across concurrent requests. Worth stating that dependency in `banner-rotation.ts`, since the file reads as safe in isolation.

**R3 (F3 residual).** `claimImpression()` latches before `trackBannerImpression` gets a call through, so a 20s timeout still loses the impression permanently with no retry. Much narrower than the 5s version, and the alternative (latch on send) reopens the double-count F1 closed, so this is probably the right trade. Worth a comment noting the state exists.

## 🎉 Praise

- `claimImpression()` returning a boolean rather than exposing the flag makes the "exactly one caller wins" contract enforceable, and the header comment reconstructs the whole resize failure so the reason the file exists is not lost.
- `IMPRESSION_READY_TRIES` vs `CLICK_READY_TRIES` as two named constants, each with a comment on why its budget differs, turns a subtle metric bias into something a reader can see and reason about.

dhananjay6561:
addressed iteration 3 in fd620ed:

r1 (hash in page key): fixed as a real bug. toc clicks go through history.replaceState('#id') directly, so asPath only picks up the fragment on a later back/forward popstate, and at that point the old code re-picked the banner and re-armed the latch, firing a second impression for the same page view. that inflates the denominator which is the whole thing this feature guards against. now keys on router.asPath.split('#')[0], so hash only changes dont even re-run the effect.

r2 (module state): added a comment noting the pick/latch is safe only because the sidebar is dynamic(..., { ssr: false }). if it ever gets server rendered that state would leak across concurrent requests, so worth calling out.

r3 (latch before send): left the behaviour as is since latching on a confirmed send would let the second sidebar instance win the race and re-open the double count f1 closed. just documented that the drop on a 20s timeout is deliberate so nobody flips it back.

tsc clean, test:unit 38/38, next lint clean.

amaan-bhati:

At `fd620ed` (1 new commit). Reviewed from the diff. **`tsc` / `test:unit` not run this round** (no disk space on the review machine), so treat CI as the source of truth for those.

## 🚦 Verdict: ✅ APPROVE / nothing open

| 🔴 0 | 🟡 0 | 🟢 0 | 💡 0 | 🎉 2 |
|---|---|---|---|---|

## ✅ Iteration 3 verified, all three closed

| # | Status |
|---|---|
| R1 page key included the `#fragment` | ✅ `router.asPath.split("#")[0]`, with the `replaceState` / `popstate` path written into the comment so the reason survives |
| R2 module state safe only under `ssr: false` | ✅ Stated explicitly, including what to do first if the sidebar is ever server-rendered |
| R3 latch set before the send is confirmed | ✅ Documented as a deliberate trade, naming the alternative and why it reopens the double-count |

That closes everything raised across the three iterations. The impression path now has one banner and one impression per page view, separate ready-budgets for clicks and impressions, the shared latch, and tests for the fire-once contract.

## 🎉 Praise

- All three items were closed the cheapest correct way: one of them a real code fix, two of them comments. Knowing which findings want code and which want a written-down reason is the part that keeps a file maintainable, and R3 in particular is now a documented decision instead of a silent one.
- The `banner-rotation.ts` header reconstructs the whole resize double-count failure. Someone reading that file cold understands why module-scoped mutable state is the right answer here, which is the thing that stops it being "cleaned up" later.

dhananjay6561:
thanks for the iteration 4 pass @amaan-bhati, appreciate you tracing the resize path all the way through across the rounds.

on the ci caveat from your side (no disk to run tsc/test:unit): all green now, build, e2e chromium, lighthouse and dco all pass, and test:unit is 38/38 in the build. so the fire-once invariant and the split ready budgets are covered by ci.

nothing open on my end, r1 was the real fix and r2/r3 stay as documented decisions so the reasoning survives in the file. ready to merge whenever a required approve lands.

amaan-bhati:
# Claude Review Skill - Iteration 5

**Head:** `fd620ed` (unchanged since Iteration 4)
**Scope:** no new commits, so this was a fresh pass rather than a delta check. Iteration 4 approved this commit on the strength of the shared-latch module; this round I traced the component's state lifecycle across an SPA navigation, which the earlier passes did not. That surfaced one measurement bug.

## 🟡 Verdict: one fix before merge

| 🔴 | 🟡 | 🟢 | 💡 |
|---|---|---|---|
| 0 | 1 | 2 | 1 |

The module (`lib/banner-rotation.ts`) is correct and I re-verified it. The bug is in `SidebarAdBanner`'s own state.

---

## 🟡 N1. `loaded` / `errored` never reset, so the new impression gate leaks across SPA navs

`components/BlogSidebar.tsx:183-184`. Both are only ever set to `true`; nothing resets them when `banner` changes. Neither `<BlogSidebar />` render site in `post-body.tsx:622,628` carries a `key`, so an SPA nav between two `[slug]` posts reconciles the same component instance and the state survives.

The state itself is pre-existing on `main`. What is new here is the gate that reads it:

```
if (!banner || hasFiredImpression() || !(loaded || errored)) return;
```

Replaying the real module against the component's state model:

```
/blog/technology/post-a: gate closed (loaded=false errored=false)   <- correct
/blog/technology/post-b: impression fired for banner_1 | painted? true
/blog/technology/post-c: impression fired for banner_1 | painted? true
final state: loaded=true errored=true
```

Two consequences:

**a. The gate is bypassed on every SPA nav.** On nav, effect 1 re-picks and re-arms the latch, then effect 2 runs immediately with `loaded` still `true` from the *previous* post. The impression is claimed before the new artwork has attempted to load. If it then fails, an impression was counted for artwork that never painted, which is the exact case the comment says the gate prevents ("a slow/blocked banner showing just the skeleton shouldn't count toward the denominator"). Internal links are the normal reading path on the blog, so this is not an edge case.

**b. One image failure sticks for the session.** Once `errored` is `true`, `if (banner && errored)` returns the text fallback on every subsequent post until a hard reload. A transient S3 hiccup on one article permanently degrades the slot. Worse for this PR's purpose: the fallback's "Sign up →" text link and the artwork's full-width "Try for Free!" button report the same `banner_click` / `banner_id`, so the dashboard silently blends two creatives with very different click rates into one CTR.

**Fix**, one effect:

```tsx
// Re-arm the paint gate when the artwork changes; otherwise `loaded` from the
// previous post opens the gate before the new banner has painted, and a single
// `errored` pins the fallback card for the rest of the session.
React.useEffect(() => { setLoaded(false); setErrored(false); }, [banner?.src]);
```

Keying on `banner?.src` rather than `banner` matters: when the random pick returns the same banner (1 in 4), the deps do not change, the effect does not run, and an already-painted image correctly keeps `loaded = true`.

---

## 🟢 Nits (2)

**N2. The observer is only disconnected by the winner.** `BlogSidebar.tsx`, impression effect: `io.disconnect()` sits inside the `&& claimImpression()` branch, so an instance that loses the claim (the second sidebar, or the one revealed by a resize across 1440px) keeps its callback attached for the page lifetime. Not a correctness problem, `claimImpression` guards the send. Disconnecting on first intersection regardless is a one-liner.

**N3. One alt string serves all four banners, with hardcoded counts.** `alt="Keploy — 300M+ mocks and 12.8M+ tests generated"` is emitted for `banner_1` through `banner_4`, whose artwork differs, and the figures go stale with no signal. Per-banner alt text on the `AD_BANNERS` entries would fix both, and the file is already the single source of truth for everything else about a banner.

---

## 💡 N4. The bug is not unit-testable in this repo, which is worth recording

`test:unit` is `tsx --test tests/lib/*.test.ts` with no jsdom or Testing Library, so the 4 module tests are the correct scope and not an oversight. But N1 lives in component state, so Playwright is the only place it can be covered: SPA-navigate between two posts and assert exactly one `banner_impression` per view, with the artwork painted at the time it fires. Worth adding alongside the fix, since this class of bug is invisible to the current suite.

---

## ✅ Re-verified this round

| Check | Result |
|---|---|
| `test:unit` | 38/38 pass |
| `tsc --noEmit` | Clean (0 errors; a bare worktree also reports missing `*.webp`/`*.png` modules, which is just the gitignored `next-env.d.ts`, not a real error) |
| Poll timeout comments | Accurate: `80 x 250ms = 20s` impressions, `20 x 250ms = 5s` clicks |
| Instance visibility | `hidden min-[1440px]:flex` / `min-[1440px]:hidden` means exactly one instance is `display:none`, so its observer never intersects and its `loading="lazy"` image never loads. The shared-latch design is sound |
| Resize across 1440px | Revealed instance loses `claimImpression()`, no double count. The original bug stays fixed |
| Iteration 3 fixes (R1/R2/R3) | Still in place |
| Branch state | 0 behind `origin/main`, 5 ahead |

The module and its tests are good work and I am not asking for changes there. N1 is a two-line effect.

dhananjay6561:
hey @amaan-bhati, dug into n1. i don't think it reproduces as a live bug, but you did surface a real fragility so i hardened it anyway.

the scenario needs loaded/errored to survive an spa nav on the same component instance. but _app.tsx renders `{loading ? <PageLoader/> : <Component/>}` and flips loading true on routeChangeStart, so the whole post subtree (including SidebarAdBanner) unmounts and remounts on every nav. loaded/errored are component useState so they start false on each post, the gate isn't bypassed and one image error doesn't pin the fallback for the session. the key you checked for on BlogSidebar isn't what's saving it, the ancestor remount is. worth noting the replay in the review is a modeled state machine (loaded stays true from the previous post), which is the exact step the remount breaks.

that said, relying on a distant ancestor to remount us is the kind of thing that breaks silently if _app ever changes, so i added the reset keyed on banner?.src anyway in 0f8dc5d as hardening, with a comment saying it's defense not a fix. now the gate is correct on its own, and your banner?.src vs banner point is right so that's what i keyed on.

nits:
- n2: left it. the losing instance is the display:none one (your own re-verify notes its observer never intersects), so disconnecting it is a no-op. happy to add the one-liner if you'd still prefer it.
- n3 (per banner alt + stale counts): fair, but the fix needs real per-banner copy and i'd rather set that deliberately with devrel than guess. can do as a follow-up.
- n4: agree, i'll add the playwright case as a follow-up to lock the one-impression-per-view behaviour in, since it's the only place it's coverable.

tsc clean, 38/38 unit still pass.

amaan-bhati (approved):
## Approving. Verified locally at `0f8dc5d`

Ran the branch in a browser rather than just reading the diff. Isolated worktree, `next dev`,
headless Chromium via the repo's own Playwright. `gtag` calls captured by pre-seeding
`window.dataLayer` with a patched `push`, so the real inline GA config reuses the array and
every event is recorded.

`npm run test:unit` 38/38, `npx tsc --noEmit` clean.

| # | Behaviour | Result |
|---|---|---|
| 1 | One impression per page view, only after the artwork paints | ✅ |
| 2 | Both sidebars agree on one banner | ✅ |
| 3 | Resize across 1440px does not double count | ✅ |
| 4 | SPA nav: one impression per view, nothing premature | ✅ |
| 5 | `#fragment` change fires nothing, banner unchanged | ✅ |
| 6 | `banner_click` carries the rendered banner's id | ✅ |
| 7 | Fallback card counts, and does not pin after recovery | ✅ |
| 8 | Warm cache does not strand the paint gate | ✅ |

nehagup (commented):
## Review — banner impression tracking (per-banner CTR)

**Recommendation: Approve with nits.** The analytics design is correct and carefully reasoned. The three things that make the CTR math valid all hold:

- **Join works** — click and impression send the identical GA4 param `banner_id` (`banner_click` / `banner_impression`, and `banner_clicked` / `banner_shown` for Clarity), so `CTR = banner_click / banner_impression` groups cleanly per banner.
- **Visibility-based** — impression fires from an `IntersectionObserver` at `threshold: 0.5`, additionally gated on `(loaded || errored)` so a skeleton-only banner doesn't inflate the denominator. Not fire-on-mount.
- **De-duplicated** — the module-scope `claimImpression()` latch fires once per page key; the observer disconnects after claiming, and the second (`display:none`) sidebar copy has a zero intersection rect so it never double-counts. The shared per-page pick guarantees both copies agree on the same `banner_id` across a 1440px resize.

SSR-safe, gtag/dataLayer/Clarity-absence safe (no-ops), observer + poller cleaned up on unmount, `pagePath` hash-stripped so a TOC `#fragment` change doesn't re-arm the latch.

### Nits
- **Latch-before-send**: `claimImpression()` sets the latch before the send is confirmed, so a fully analytics-blocked client silently drops the impression with no retry. It's a deliberate trade to prevent the second instance re-opening the double-count; net bias is toward *understating* CTR, never inflating — acceptable, just documenting it.
- `whenReady` leaves its `abort` listener registered on the success path (micro-leak; `once:true`, negligible).
- `threshold: 0.5` assumes the banner is shorter than the viewport — fine for the current slot; revisit if the artwork ever grows taller than a small mobile viewport.
- CI tests pin the pure rotation/latch module (the highest-risk logic) but not the visibility gate / `loaded||errored` / `pagePath` keying — a jsdom test there would protect the parts most likely to regress.

nehagup:
You may ignore this review.

nehagup (approved):

dhananjay6561:
thanks @nehagup

PR #418, opened 2026-09-02
feat(blog): add summarize with ai deep links on blog posts

dhananjay6561:
## what this does

adds a small "summarize this blog post with" row at the top of every post, right under the author block. its just icon buttons, no labels:

- chatgpt
- google ai
- perplexity
- claude
- grok

clicking one opens that assistant with a prompt asking it to read, summarize and cite the current post, and follow the internal links inside it.

## why

mainly for ai citation and indexing:

- makes it easy for readers to pull the post into an ai tool
- nudges those tools to ingest the post and surface keploy as a cited source
- gets the internal links crawled along with it

## details

- **google ai** points at googles ai mode search (`udm=50`) so the prompt actually lands there instead of a plain search
- the rest open their chat uis with the prompt prefilled
- the prompt is a single line on purpose, long or multi line queries make google ai mode drop it and fall back to a normal search
- the brand icons are webp files hosted on s3 (`keploy-devrel`) and loaded as plain img tags, colours are baked into the webp so no per tool styling is needed and we dont add react icons as a dependency
- the widget lives in the shared post header so it shows on both technology and community posts

## testing

- ran locally on a technology post, all five buttons render from s3 and deep link correctly
- layout wraps fine on mobile
- `tsc` passes


dhananjay6561:
good catch, fixed. the gradient id is now generated per instance with react useId so it cant collide even if the icon renders more than once. pushed in 6d6743d.

amaan-bhati (approved):
LGTM

Tested locally on desktop and smaller devices as well:

<br/>

Image

<br/>


<br/>

Image


The links are also opening with the prompts, verified locally:
<br/>

Image

<br/>

<br/>

Image

<br/>


Image

<br/>

nehagup (approved):

PR #419, opened 2026-09-04
feat(blog): add preferred source on Google button to post header

dhananjay6561:
three blog changes in here.

## 1. add as preferred source on google button

small "add as preferred source" button in the post header, right next to the summarize with ai icons. opens google's preferred sources picker prefilled with keploy.io, so when a reader adds keploy, google shows our posts more in their top stories and search.

- new PreferredSourceButton component, white pill with the google g mark and orange hover so it matches the ai icons
- sits on the same wrapping row as the summarize icons, drops to its own line on narrow screens, no overflow down to 320px

## 2. fix silent newsletter subscribe failures

filling all three fields and hitting subscribe sometimes did nothing, no success and no error. handleSubscribe was doing `response.data.subscription`, which throws when the response has no `data` (missing url env, or a `{ errors: [...] }` shape from the backend), and it got swallowed so the user saw nothing.

- optional chained the success check so it doesnt crash on undefined or error shaped responses
- `subscribeMutation` now rejects clearly when the url env is missing instead of returning undefined and blowing up the caller
- abortcontroller timeout on the subscribe fetch so a stuck backend cant leave the button disabled forever
- visible "something went wrong, please try again" on failure
- on success the whole form gets swapped for a "thank you for subscribing" panel so its obvious it actually worked (was a tiny green line for 3s before)

## 3. json-ld schema validator + reviewedBy warning fix

ran the live post pages through validator.schema.org and every post throws one warning: `reviewedBy` was being emitted on the BlogPosting/TechArticle node, but in schema.org `reviewedBy` is a webpage only property. so the site was 0 errors but 490 warnings across all posts (455 blogposting + 35 techarticle), all the same thing.

two parts:

- **fix** moved `reviewedBy` onto the webpage that `mainEntityOfPage` already points at, instead of the article node. keeps the reviewer eeat signal, just on a node schema.org actually allows it on. no copy or ui change.
- **guard** added `scripts/validate-schema.ts` (`npm run validate:schema`) so this class of bug gets caught before deploy instead of clicking every page through validator.schema.org by hand. it sweeps the rendered routes, pulls every ld+json block and lints for wrong property for a type, invalid enum values, malformed json, duplicate @id, and a top level node with no @type. ported from the landing validator (keploy/landing#286) and adapted for this repo (pages router, basePath /blog, routes discovered from public/sitemap.xml). rules are an extensible `(nodes) => findings[]` list, the pure extract/lint functions are unit tested via the existing node:test runner, and `--self-test` runs fixtures with no server needed.

how it ran:
- `npm run test:unit` passing, added the validator tests plus a pin that feeds the builder output through the linter so `reviewedBy` can never drift back onto the article node
- full prod sweep before the fix: 0 errors, 490 `reviewedBy` warnings (this is what the fix clears once deployed)
- local dev sweep with the fix (`SCHEMA_VALIDATE_BASE_URL=http://localhost:3000 npm run validate:schema`): the same sampled routes went from 6 warnings to **0 errors / 0 warnings**

## notes

stacked on #418 (the button lives inside the summarize row), so this targets main and will show #418s commits until that merges, then it drops to just these files. newsletter fix verified live with playwright: success shows the thank you panel, failure shows the error, no crash either way. schema validator is script + tests only, no runtime or ui change; the reviewedBy fix is a schema shape change only.


components/subscribe-newsletter.tsx
```diff
@@ -69,20 +84,23 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean })
   const handleSubscribe = async (payload) => {
     try {
       const response = await subscribeMutation(payload);
-      if (response.data.subscription) {
+      // Optional-chain the whole path: `response` is undefined on a
+      // misconfigured URL and `response.data` is absent on a GraphQL error
+      // shape ({ errors: [...] }) — both used to throw "cannot read 'data'"
+      // here and get swallowed, leaving the user with no feedback.
+      if (response?.data?.subscription) {
```
dhananjay6561, line 91: quick sanity check on the success detection. it treats `response.data.subscription` as truthy = success, but if the graphql resolver can return a legit falsy value (say `false` for an email thats already subscribed, or an empty string) the user lands in the else and sees 'something went wrong' even though the mutation actually ran fine. do we know what `subscription` resolves to on a real success? if it can be falsy it might be safer to key off `response?.data && !response?.errors` rather than the field value. not blocking if the backend always returns true.

scripts/validate-schema.ts
```diff
@@ -0,0 +1,490 @@
...
+
+  let errorCount = 0;
+  let warningCount = 0;
+  let unreachable = 0;
+
+  for (const route of routes) {
+    const url = `${BASE_URL}${route}`;
+    let html: string;
+    try {
+      const res = await fetch(url);
```
dhananjay6561, line 440: nit, the sweep awaits each fetch one at a time in the loop. totally fine for the sampled default, but with --all over the whole sitemap (1600ish routes) this runs fully serial and will be slow. if it ever gets wired into ci on the full set itd be worth a small bounded concurrency pool. not blocking.

dhananjay6561 (commented):
reviewed with the awesome-skills/code-review skill (typescript + react + common bugs guides). solid pr, all three changes hang together well. the newsletter fix is defensive in the right places (optional chaining the success path, the abortcontroller timeout so a hung backend cant strand the button, swapping the whole form for the thank you panel), and the schema #3 change is a clean minimal move of reviewedBy onto the webpage node which is exactly where schema.org wants it. the validate-schema script with the ported rules + self test is a nice guardrail.

left two small inline notes below. one extra type nit that didnt land on a changed line: handleSubscribe's `payload`, submitHandler's `e`, and isValidEmail's `email` all come through as implicit any, so the form flow loses type safety. could type payload as the shape subscribeMutation already declares and `e` as React.FormEvent<HTMLFormElement>. purely a nit, not blocking.

components/subscribe-newsletter.tsx
dhananjay6561: dug into this and actually reproduced it locally. spun up a mock returning {data:{subscription:false}} with a 200 and no errors, and the form drops straight into 'something went wrong' even though the mutation ran fine, so the concern is real on our side. the open bit is whether the live backend ever returns a falsy subscription (say false for an already subscribed email). confirming that with the backend, and if it can then ill switch the check to keying off response?.data && !response?.errors. verified {data:{subscription:true}} still shows the thank you panel correctly.

components/subscribe-newsletter.tsx
```diff
@@ -223,10 +255,11 @@ export default function SubscribeNewsletter(props: { isSmallScreen?: boolean })
               *<strong>We won&#39;t spam you</strong> only one Email every month.
             </span>
             <RecaptchaAttribution />
+            {emailError && <p className="text-sm text-red-500 text-center font-semibold mt-1">{emailError}</p>}
+            {submitError && <p className="text-sm text-red-500 text-center font-semibold mt-1">{submitError}</p>}
           </div>
+          )}
           {recaptchaScript}
```
dhananjay6561, line 265: small a11y thing on the feedback messages. both emailError and submitError render as plain `<p>`s, so a screen reader wont announce them when they pop in (a sighted user sees the red text, a screen reader user just gets a button that silently re enables). wrapping these in `role="alert"` (or `aria-live="assertive"`) would announce the failure. same idea for the 'thank you for subscribing' panel above, a `role="status"` there would announce the success too. not blocking.

dhananjay6561 (commented):
took a second pass, this time against the full pr diff on github (my first pass had a stale local base so it surfaced some already merged structured-data code, please ignore that part). the three real changes hold up well:

- newsletter: subscribeMutation is only called from this file and that one caller catches, so switching missing url from a silent undefined to a rejected promise is safe. the abortcontroller timeout + visible submitError + swapping the form for the thank you panel all read cleanly.
- schema #3: the reviewedBy move onto the mainEntityOfPage webpage node is a clean minimal fix, and mutating mainEntityOfPage after it's embedded in schema works fine via reference.
- validate-schema: the pure lintBlocks + node:test coverage + self-test is a good guardrail.

my two earlier notes still stand (the response.data.subscription truthiness check, and the serial fetch under --all). one new small a11y note below, otherwise looks good to me.

components/subscribe-newsletter.tsx
dhananjay6561: good catch, done in 08588e2. added role="alert" to both emailError and submitError so the failure gets announced when it pops in, and role="status" on the thank you panel so the success is announced too. kept them as the assertive vs polite split you suggested.

scripts/validate-schema.ts
```diff
@@ -0,0 +1,490 @@
...
+      (unreachable ? `, ${unreachable} unreachable` : ""),
+  );
+  if (unreachable === routes.length) {
+    console.log(
+      `\nAll routes were unreachable — is the app running at ${BASE_URL}? ` +
+        `Start it with "npm run dev" or set SCHEMA_VALIDATE_BASE_URL.`,
+    );
+    return 1;
+  }
+  return errorCount > 0 ? 1 : 0;
```
dhananjay6561, line 479: worth a decision on the exit code here. the run only fails hard when ALL routes are unreachable (the check above), so a partial failure like 3 of 10 post routes returning 404/500 just prints SKIP and, as long as the reachable ones are clean, the script still exits 0. if this ever gates ci that means a batch of broken or missing pages goes green under a 'no schema errors' banner. might be worth treating unreachable > 0 (or over some threshold) as a failure too, or returning a distinct non zero code, so a page that quietly stopped rendering doesnt slip through. fine to leave if it's only ever run by hand.

dhananjay6561 (commented):
third pass, focused on the tests + the script's ci behaviour this time.

tests look great: validateSchema.test.ts covers every rule including both duplicate-id cases and the @type array union, and the structuredData test pins reviewedBy onto the webpage AND runs it back through lintBlocks end to end so it cant regress. nice. also confirmed tsx is already a devDependency (test:unit uses it) so the new validate:schema script runs fine, and importing lintBlocks is side effect free thanks to the require.main guard.

one design question on the exit code below. nothing blocking.

scripts/validate-schema.ts
dhananjay6561: fair point. leaving it as is for now since the script isnt wired into ci or any hook yet, its only run by hand, so the partial unreachable case has no consumer relying on the exit code today. if we ever gate ci or a pre-push hook on it ill treat unreachable > 0 as a failure too (or a distinct non zero code) so a page that quietly stopped rendering cant slip through green. noting it here so that decision isnt lost.

components/subscribe-newsletter.tsx
dhananjay6561: pulled 08588e2, confirmed. role=alert on both error <p>s and role=status on the thank you panel, and since those roles imply assertive vs polite live regions respectively you get the split for free without needing explicit aria-live. resolving this one, thanks for the quick turnaround.

components/subscribe-newsletter.tsx
dhananjay6561: nice repro, that {data:{subscription:false}} + 200 + no errors case is exactly the gap. one thought while you confirm with backend: even if it returns true today, keying off response?.data && !response?.errors is strictly safer since it treats 'mutation ran, no graphql errors' as success rather than trusting one field's truthiness, so it wont silently break if the resolver's return shape ever shifts. fine to leave pending their answer though, your call.

scripts/validate-schema.ts
dhananjay6561: sounds good, deferring is reasonable while nothing consumes the exit code. the in thread note is enough to not lose it. if/when it gets wired into ci or a pre-push hook, unreachable > 0 as a fail (or a distinct non zero code) is the thing to add. all good.

dhananjay6561 (commented):
lgtm from my end. reviewed all three changes end to end, the newsletter fix, the reviewedBy webpage move, and the schema validator + tests all hold up, and the a11y fix in 08588e2 is verified (role=alert on the errors, role=status on the thank you panel). (cant hit the formal approve button since im the pr author, but consider this a lgtm.)

two non blocking follow ups already tracked in the threads so they dont get lost:
- line 91, the response.data.subscription success check is a confirmed real bug for a falsy but successful response. youre checking the backends return shape and will switch to response?.data && !response?.errors if it can return falsy. fine as a follow up.
- validate-schema partial unreachable exit code, consciously deferred until it gates ci or a hook.

nice work, the defensive error handling and the ported schema linter + tests are a solid guardrail.

amaan-bhati (approved):
Reviewed locally on `feat/preferred-source-google` (`832b3fa`). All three changes work as described.

<br/>
Image
<br/>

<br/>
Image
<br/>


**Verdict:** Code is correct - happy with all three on their merits. Please split the newsletter fix into its own PR, and land #418 first (this is stacked on it). Rest are follow-ups.

### Ran locally
- `npm run test:unit` - 121/121 pass
- `validate-schema --self-test` - 7/7 pass
- Full sweep vs localhost - 13 routes, **0 errors / 0 warnings**
- `tsc --noEmit` clean, no new lint warnings
- Button renders in post header, href resolves (200)
- Newsletter: with the subscription URL unset locally, failure now shows "Something went wrong" instead of the old silent dead button

Checked `reviewedBy` **moved** rather than got dropped - "0 warnings" is also what deleting it would produce. Reviewer (`Neha Gupta`) is correctly nested on the `WebPage` node on three sampled posts, `BlogPosting` + `TechArticle`. E-E-A-T intact.

### Blocking
**1. Split the PR** - 976 lines, three unrelated changes, three risk profiles. If the newsletter fix breaks prod, reverting it also reverts the schema fix and the validator.

### Follow-ups
**2. `validate:schema` isn't in CI** - not referenced anywhere in `.github/`. The `reviewedBy` pin *is* covered (`test:unit` runs in `playwright.yml`); it's the route sweep that's manual. It needs a running server, but `playwright.yml` already boots one. Either wire it up or soften the description.

**3. The `lintBlocks` assertion in the new test isn't load-bearing** - `collectNodes` doesn't recurse into property values, so the now-nested `reviewedBy` is never inspected; that assert would pass even with `reviewedBy` on a nested Article. The `!("reviewedBy" in post)` line above it is the real pin. Drop the `lintBlocks` half or recurse.

**4. 8s abort can report failure on a subscribe that succeeded** - aborting client-side doesn't stop the server. Retry re-runs `submitHandler` and re-fires `sendBlogLead` + `/api/blog-lead-notify` → duplicate lead + Chat ping. `submittingRef` guards concurrent clicks, not sequential retries. 15s would cause fewer false failures.

**5. `#ED5D0F` hardcoded** - fourth occurrence (`BlogSidebar.tsx:261`, three in `lib/banner-rotation.ts`, now here). Belongs in `tailwind.config.js`.

**6.** No click tracking on the button (the point is measurable lift), and Preferred Sources is region/sign-in gated so it's inert for some traffic.

### Nice work
Accessibility on the newsletter change is solid — `role="status"`, `role="alert"`, `aria-hidden` on decorative SVGs, real `aria-label`. `Promise.reject` over `undefined` is right; `subscribeMutation` has no other callers. `require.main === module` guard is correct too.

amaan-bhati:
@dhananjay6561 Also please resolve the conflicts as well.

PR #420, opened 2026-09-04
fix(blog): stop silent newsletter subscribe failures, show clear result

dhananjay6561:
## what

filling all three fields and hitting subscribe sometimes did nothing. no success msg, no error, button just felt dead.

## why

handleSubscribe was doing `response.data.subscription`. on any non success response that value isnt there:

- `subscribeMutation` returns `undefined` when `NEXT_PUBLIC_NEWS_LETTER_SUBSCRIPTION_URL` isnt set, so it breaks every single time on a deploy thats missing the env
- the backend sends back a `{ errors: [...] }` / `data: null` shape on a bad or flaky response, thats the "sometimes" in prod

both cases make `response.data.subscription` throw "cannot read properties of undefined (reading 'data')", which got caught and only console.error'd, so the user saw nothing at all. repro'd it live on the post page form.

## what i changed

- optional chained the success check (`response?.data?.subscription`) so it doesnt crash on undefined or error shaped responses
- `subscribeMutation` now rejects with a clear error when the url env is missing, instead of returning undefined and blowing up the caller downstream
- added an abortcontroller timeout on the subscribe fetch so a stuck backend cant leave the button disabled forever
- show a visible "something went wrong, please try again" on failure
- on success, swap the whole form out for a "thank you for subscribing" panel so its obvious it actually worked (before it was a tiny green line for 3s that was super easy to miss)

## testing

drove the real form with playwright, tsc + eslint clean:

- success (`{data:{subscription:true}}`) shows the thank you panel, no error, no console errors
- failure (`{errors:[...]}`) shows the error, no console errors
- the old cannot read 'data' typeerror is gone in both paths

### DIFF
--- a/components/subscribe-newsletter.tsx
+++ b/components/subscribe-newsletter.tsx
@@ -7,34 +7,46 @@ import { sendBlogLead } from "../lib/blog-mql";
 
 export const subscribeMutation = (formData: { fullName: string, email: string, companyName: string, message: string }) => {
 
-  if (newsLetterSubscriptionUrl){
-    return fetch(newsLetterSubscriptionUrl, {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-        'Origin': 'https://keploy.io',
-      },
-      body: JSON.stringify({
-        query: 'mutation Subscribe($input: GuestInput!) { subscription(guestInput: $input) }',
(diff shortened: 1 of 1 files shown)


dhananjay6561:
folding this into #419 instead

